Issue #172, opened 2015-02-11
Double Ctrl-O required to go back on some sites


sdothum:
Some sites seem to resist going back.. but I found that if I do the Ctrl-O twice in sequence, relatively quickly, vimb does go back. E.g.

http://imgur.com/a/d9Zvp

from 

http://www.reddit.com/r/unixporn/comments/2vhn1m/what_are_your_favorite_beautiful_applications/ 

(the comment link "here" by redditErick to the imgur web page)

Pressing a single Ctrl-O does not go back. A quick double Ctrl-O does go back.

fanglingsu:
I can reproduce this. On a first try I thought this is related to JavaScript. Because if JavaScript is disabled the click of the 'here' link leads to the imgur url and this redirects to a noscript variant of the page. The `<C-O>` brings vimb back to the imgur url which redirects immediately back to the noscript url http://imgur.com/a/d9Zvp/noscript. To go back I've to press `<C-O>` three times fast.
But the issue appears also if JavaScript is enabled, even if vimb goes back after two `<C-O>`.

fanglingsu:
This issue seems to be related to HSTS (HTTP Strict Tranport Security) feature of vimb. If `:set hsts=off` the `<C-O>` leads direct back to the previous page.

sdothum:
Thanks Daniel! I can live with hsts disabled for now while you ponder the problem.. as there are enough sites whose url I don't remember which present this history navigation probleem.

semarie:
@fanglingsu I have followed your idea to reload the frame only if it is the main request, with some adaptations.

A test against `soup_message_get_first_party` would work with some code (it is the request as requested by user, so with "http://") So, in order to keep that simple, the code reload the page only if it is the main-frame (we have `view` and `frame` here: so just checking if `frame` is the mainframe is ok).

I have do some tests: it corrects #172, and seems to be ok for others cases (HSTS in simple case or after redirects). Please do some tests too: HSTS is a security feature that we shouldn't break :-)

fanglingsu:
@semarie The two issue are solved. I've done some tests with included sources like images, and the url are properly rewritten to https (tested with wireshark).

fanglingsu:
In case I set the `ca-bundle` to an empty file the hsts prevents loading of the https resources like expected, independent from the `strict-ssl` option.

shioyama:
[...]

Issue #146, opened 2014-11-15
hsts / redirect problem


semarie:
Hi Daniel,

I encounter an issue with hsts: it don't catch all possibles cases for url rewrite, and it is possible to by-pass it.

Here an example:
- first, visit https://www.eff.org/ in order to see the HSTS header one time at least.
- now, if you go to http://www.eff.org/ you will be redirected to https version
- now, go to http://bit.ly/1zZ4fzo it is a shorten url for http://www.eff.org/

vimb don't redirect to https version, and you arrive on the http version.

fanglingsu:
@semarie Our hacky solution creates additional entries in webkits back/forward history list like mentioned in referenced issue.

semarie:
I'm not sure to be able to provide alternate solution which works for HSTS... a possible workaround would be to have a (temporary) setting for configure if HSTS should be enable for redirect, and if setted, which don't call the `webkit_web_frame_load_uri` part. But it would complexify the vimb code.

fanglingsu:
I'm not sure, but maybe we can identify the main request that load the entire pages html by using [soup_message_get_first_party](https://developer.gnome.org/libsoup/stable/SoupMessage.html#soup-message-get-first-party). If this would be possible, we could do the `webkit_web_frame_load_uri`, else the requeueing of the soup request should be sufficient.

But might be that the first party stuff is also true for pages loaded in (i)Frames, which is normally the case for social media links.

If you have some spare time, I'll invite you to check if this works.

I suggest to do further discussions in #172.

PR #177, opened 2015-02-13
hsts: reload only the main frame

semarie:
HSTS in done using requeueing. But due to a bug in webkit, a reload of
the frame could be required in order to display the redirected URI (and
not the original one).

The patch force this reload only for the mainframe (where the correct
URI is need), and not for the subframe (where the HSTS stuff is done by
requeueing).

Original idea from @fanglingsu 

This patch follow the discussion in #146 and should address #172.


fanglingsu:
Wow, pretty nice, that works fine. Thank you for this patch.

Issue #184, opened 2015-03-08
Option to remove auto-triggering hinting


mynameisfiber:
I really love the hinting system, but sometimes when I'm typing out the hint and it auto-triggers because there is only one thing left, the page changes but I'm still typing and then random commands get triggered!  I would much rather be able to change this functionality so I _always_ have to press enter to continue.

mynameisfiber:
For some reason this segfaults when set in the config file, however setting during runtime with `:set hint-follow-last=false` works great.  This seems to be a general problem, however... any configuration in `$XDG_CONFIG_HOME/vimb/config` leads to a segfault.

mynameisfiber:
Ah, nevermind.  The documentation seemed to maheke me think that there was a special config format of the form,

```
key  value
```

However, trudging through the code shows that it is simply a list of commands such that the correct thing to have in the config file is,

```
set key=value 
```

Anyways... this pull is ready to go!

fanglingsu:
I've merged you patches, so I think this issue is solved by you great work.

PR #186, opened 2015-03-08
Added hint-follow-last option

mynameisfiber:
Trying to resolve issue #184, I added a new option called `hint-follow-last` that will toggle the behaviour of vimb to automatically follow the last remaining hint.  This behaviour is by default turned to `hint-follow-last=True` to continue with expected behaviour.

### DIFF
--- a/src/hints.c
+++ b/src/hints.c
@@ -162,8 +162,9 @@ void hints_create(const char *input)
             JSValueMakeBoolean(hints.ctx, hints.gmode),
             JSValueMakeNumber(hints.ctx, MAXIMUM_HINTS),
             js_string_to_ref(hints.ctx, GET_CHAR("hintkeys")),
+            JSValueMakeBoolean(hints.ctx, GET_BOOL("hint-follow-last"))
         };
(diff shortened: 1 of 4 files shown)


mynameisfiber:
For some reason this segfaults when set in the config file, however setting during runtime with `:set hint-follow-last=false` works great.  This seems to be a general problem, however... any configuration in `$XDG_CONFIG_HOME/vimb/config` leads to a segfault.

mynameisfiber:
Ah, nevermind.  The documentation seemed to maheke me think that there was a special config format of the form,

```
key  value
```

However, trudging through the code shows that it is simply a list of commands such that the correct thing to have in the config file is,

```
set key=value 
```

Anyways... this pull is ready to go!

fanglingsu:
Thank you for the patches. I've squashed your commits into a single one and merged it into master.
I'm not so satified with this solution, but I've no other really usable idea on how to avoid to run normal mode commands on the attempt to filter hints. The only alternative I see would be to eat all keypresses for some milli seconds after a hints was fired automatically.

PR #190, opened 2015-03-13
New homepage feature

Artymort:
Hi,

I quite like the usual idea of a homepage with quick access to your bookmarks. This is not exactly useful in the case of vimb, but I still like it. So I added it =D

My C is a bit rusty, I tried to follow your conventions but maybe there are major parts that need to be modified. Actually I wrote this in a separate project before thinking about merging it into ```bookmark.c```, so maybe there are function to parse bookmarks file that I could/should have used. That said it works and seems stable.

The new behavior is the following one :
* If there is no bookmark, the homepage is the main site of vimb (the previous behavior). This makes sense as it is a manual, so as long as you don't know how to use vimb you want to access this.
* If there are bookmarks, vimb generates a lightweight html file with them and opens it.

I also added a small CSS file.

For example the following bookmark file:
```
http://www.duck.com/	Google o_O !	search engine surprise
http://www.duckduckgo.com/	DuckDuckGo	search engine noSurprise
https://github.com/fanglingsu/vimb/	Github - vimb	source vimb
```

gives a page that looks like this: 
bookmark

Of course changing the homepage in the config file override this feature.

Thank you for your work, I hope this suits you !

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ install: $(TARGET) $(DOCDIR)/$(MAN1)
 	install -d $(DESTDIR)$(EXAMPLEDIR)
 	cp -r examples/* $(DESTDIR)$(EXAMPLEDIR)
 	install -d $(DESTDIR)$(MANDIR)/man1
+	cp $(SRCDIR)/bookmark.css $(DESTDIR)$(SHAREDIR)
 	@sed -e "s!VERSION!$(VERSION)!g" \
 		-e "s!PREFIX!$(PREFIX)!g" \
 		-e "s!DATE!`date +'%m %Y'`!g" $(DOCDIR)/$(MAN1) > $(DESTDIR)$(MANDIR)/man1/$(MAN1)
@@ -30,7 +31,7 @@ install: $(TARGET) $(DOCDIR)/$(MAN1)
 uninstall:
 	$(RM) $(DESTDIR)$(BINDIR)/$(TARGET)
 	$(RM) $(DESTDIR)$(MANDIR)/man1/$(MAN1)
-	$(RM) -r $(DESTDIR)$(EXAMPLEDIR)
+	$(RM) -r $(SHAREDIR)
 
 dist: dist-clean
 	@echo "Creating tarball."
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ install: $(TARGET) $(DOCDIR)/$(MAN1)
 uninstall:
 	$(RM) $(DESTDIR)$(BINDIR)/$(TARGET)
 	$(RM) $(DESTDIR)$(MANDIR)/man1/$(MAN1)
-	$(RM) -r $(SHAREDIR)
+	$(RM) -r $(DESTDIR)$(SHAREDIR)
 
 dist: dist-clean
 	@echo "Creating tarball."
--- a/src/bookmark.c
+++ b/src/bookmark.c
@@ -186,6 +186,117 @@ gboolean bookmark_fill_tag_completion(GtkListStore *store, const char *input)
     return found;
 }
 
+/**
+  * Converts bookmark file to a simple HTML5 page
+  *
+  * @input_bookmark_path: the path to the input bookmark file
+  * @output_html_path: the path to the output html file
+  *
+  * Returns: A boolean to say weather there was an error or not
+  */
+int bookmark_to_html(char* input_bookmark_path, char* output_html_path)
+{
+  FILE* input = fopen(input_bookmark_path, "r");
+  FILE* output = fopen(output_html_path, "w");
+
+  if(input == NULL)
+  {
+    printf("Could not open bookmark file for reading.\n");
+    return 1;
+  }
+  else if (output == NULL)
+  {
+    printf("Could not open html file to write into.\n");
+    return 2;
+  }
+  else
+  {
+    char header[] =
+      "<!doctype html>\n"
+      "<html lang='en'>\n"
+      "\t<head>\n"
+      "\t\t<meta charset='utf-8'>\n"
+      "\t\t<link rel='stylesheet' type='text/css' href='./bookmark.css'>\n"
+      "\t\t<title>Bookmarks</title>\n"
+      "\t</head>\n"
+      "\t<body>\n"
+      "\t\t<table>\n"
+      "\t\t\t<tr>\n"
+      "\t\t\t\t<th>\n"
+      "\t\t\t\t\tBookmark\n"
+      "\t\t\t\t</th>\n"
+      "\t\t\t\t<th>\n"
+      "\t\t\t\t\tTag(s)\n"
+      "\t\t\t\t</th>\n"
+      "\t\t\t</tr>\n";
+    char footer[] =
+      "\t\t</table>\n"
+      "\t</body>\n"
+      "</html>\n";
+
+    //Can be hardcoded to avoid string.h dependency.
+    int header_size = strlen(header), footer_size = strlen(footer);
+
+    char line[512];
+
+    if (header_size != fwrite(header, sizeof(char), header_size, output))
+    {
+      printf("Something went wrong while writing header.\n");
+      return 3;
+    }
+
+    while (fgets(line, sizeof(line), input))
+    {
+      char* item1 = NULL;
+      char* item2 = NULL;
+      char* item3 = NULL;
+
+      //Tried this on ugly/malformed strings, seems OK
+      //NOT sure though ! Valgrind seems unhappy with it :/
+      item1 = strtok(line, "\t");
+      item2 = strtok(line, "\t");
+      item3 = strtok(line, "\t");
+
+      //Remove trailing \n if any
+      if (item3 == NULL && item2 != NULL)
+         item2[strlen(item2) - 1] = '\0';
+      if(item3 != NULL)
+        item3[strlen(item3) - 1] = '\0';
+
+      if(item1 != NULL && item2 != NULL)
+      {
+        fprintf(output,
+          "\t\t\t<tr>\n"
+          "\t\t\t\t<td>\n"
+          "\t\t\t\t\t<a href='%s' title='%s'>\n"
+          "\t\t\t\t\t\t%s\n"
+          "\t\t\t\t\t</a>\n"
+          "\t\t\t\t</td>\n"
+          "\t\t\t\t<td>\n"
+          "%s%s%s"
+          "\t\t\t\t</td>\n"
+          "\t\t\t</tr>\n",
+          item1,
+          item2,
+          item2,
+          (item3 != NULL) ? "\t\t\t\t\t" : "",
+          (item3 != NULL) ? item3 : "",
+          (item3 != NULL) ? "\n" : "");
+      }
+    }
+
+    if (footer_size != fwrite(footer, sizeof(char), footer_size, output))
+    {
+      printf("Something went wrong while writing footer.\n");
+      return 4;
+    }
+
+    fclose(input);
+    fclose(output);
+    return 0;
+  }
+}
+
 #ifdef FEATURE_QUEUE
 /**
  * Push a uri to the end of the queue.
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -30,7 +31,7 @@ install: $(TARGET) $(DOCDIR)/$(MAN1)
 uninstall:
 	$(RM) $(DESTDIR)$(BINDIR)/$(TARGET)
 	$(RM) $(DESTDIR)$(MANDIR)/man1/$(MAN1)
-	$(RM) -r $(DESTDIR)$(EXAMPLEDIR)
+	$(RM) -r $(SHAREDIR)
```
semarie: shouldn't be `$(DESTDIR)$(SHAREDIR)` instead of `$(SHAREDIR)` ?

src/bookmark.c
```diff
@@ -186,6 +193,145 @@ gboolean bookmark_fill_tag_completion(GtkListStore *store, const char *input)
...
+
+  //If the bookmark file is empty we redirect the user to main vimb page
+  fseek(input, 0, SEEK_END);
+  if (ftell(input) == 0)
+  {
+    fprintf(output,
+      "<!DOCTYPE HTML>\n"
+      "<html lang='en-US'>\n"
+      "\t<head>\n"
+      "\t\t<meta http-equiv='refresh' content='0; url=http://fanglingsu.github.io/vimb/' />\n"
```
semarie: maybe you should keep `SETTING_HOME_PAGE` definition, and use it here

src/bookmark.h
```diff
@@ -24,6 +24,7 @@ gboolean bookmark_add(const char *uri, const char *title, const char *tags);
 gboolean bookmark_remove(const char *uri);
 gboolean bookmark_fill_completion(GtkListStore *store, const char *input);
 gboolean bookmark_fill_tag_completion(GtkListStore *store, const char *input);
+gboolean build_html_file();
```
semarie: I think you should keep `bookmark_` prefix for exported function. It would be more evident where the function lives, and what it should do.

Makefile
Artymort: Yes, you're right. How does it work, do I change this, cancel this PR and create another one, or is there a cleaner solution ?

src/bookmark.h
Artymort: I agree with this too, same question as on the first comment.

src/bookmark.c
Artymort: I agree with this one too, thanks !

Makefile
semarie: you could "update" your PR, by updating your branch:
 - push a new patch
 - amend/rebase your patchs and do a push with `--force` option

Artymort:
Thanks @semarie for the suggestions, I patched it !

fanglingsu:
There are some parts in the code that I don't like and that can be replaced by glib functions.
```c
input_path = Emalloc((strlen(path) + 10) * sizeof(char));
memcpy(input_path, path, strlen(path));
memcpy(input_path+strlen(path), "/bookmark\0", 10);

/* this could be written as */
input_path = g_strconcat(path, "/bookmark", NULL);
/* or *
input_path = g_build_filename(path, "bookmark", NULL);
/* or best solution - use the already available bookmark file path */
input_path = vb.files[FILES_BOOKMARK];
```

But that could be changed easily. More important for me - "what's the benefit
of this compared with current implementaiton?". That are 189 additional lines
of code for somethings simple like showing the bookmark file as html.

1. You can write a little screll script that geneates the html for you and
   start vimb like `generateBookmarkHtml | vimb -`. I know that this works
   only on startup and you can't go back to the html. But you could also run
   the html generator and write the html to a known file path and configure
   this as home-page. If you check for inotify events you can always keep the
   html file up to date.
2. You could add some little JavaScript to the scripts.js file that converts
   the bookmark file to html and setup the home-page to the bookmarkfile and
   add an autocmd that runs the JavaScript function.
   ```javascript
    function bmshow() {
        var html='' ,parts=[], line, uri, lines = document.body.innerText.split(/\n/);
        for (var i = 0; i < lines.length; i++) {
            line  = lines[i];
            parts = line.split("\t");
            uri   = parts[0];
            html  = html + "<tr><td><a href=\"" + uri + "\">" + uri + "</a></td><tr>";
        }
        html = "<table>" + html + "</table>";
        document.body.innerHTML = html;
    }
    ```

    You can run this from within vimb by `:e! bmshow()`. Yes, this is an
    really simple and stupid example, but it demonstrates what I mean.

I don't like to bloat vimb by features that only a view people need and that
can already be achived from outside of vimb.

Artymort:
OK, I understand why you wouldn't want to merge this. And the JS solution is interesting, though I don't quite get what you mean by autocmd (and I know very few JS) ?

The dirty part of the code (not using ```g_strconcat``` for exemple) comes from the fact that I don't have a deep knowledge of C library. Though this is little work to patch, and doesn't change the aspect of the code, a consequent part of which is juste pretty-printing that could be skipped/one-lined.

Originately I wrote the convertion function as an external tool as suggested in 1, but then I prefer something that would update in real-time, and since I wrote it in plain C I thought that I could just integrate it in the main code.

What's more I don't think it does actually bloating vimb, with a few more line you can disable it completely for example if the homepage is present in the settings.

Eventually I'm going to try and implement the JS version, but I think for the moment I'm just going to stick to this.

Thanks for the answer !

fanglingsu:
I'd suggest you to write a script or a standalone tool that reads the bookmark file and generates the html. Then you can use http://linux.die.net/man/1/inotifywatch to observes changes on the bookmarkfile and rebuilt the html file. The everithing you need is to `:set home-page=file:///home/...`

Artymort:
Thanks for the advice. I guess I can close the PR then !

fanglingsu:
> OK, I understand why you wouldn't want to merge this. And the JS solution is interesting, though I don't quite get what you mean by autocmd (and I know very few JS) ?

You can setup autocmd like in vim that do various things on some events, depending wich URL is opened (http://fanglingsu.github.io/vimb/man.html#Automatic_commands).

```
autocmd LoadFinished file:///home/*/.config/vimb/bookmark e! bmshow();
```

fanglingsu:
If you have a nice script, independent if JavaScript or sh please feel free to provide it. So I can publish it on http://fanglingsu.github.io/vimb/scripts.html and/or put it into the example directory of the project so that other user that like the approach can use this.

Artymort:
I wrote a small sh script that does what I wanted, it uses inotify (thanks). I wouldn't say I tested it very deeply but it seems to be working fine.

The code is [here](https://github.com/Artymort/vimb/blob/master/src/vimb_bookmark) and the css I use is [there](https://github.com/Artymort/vimb/blob/master/src/bookmark.css), you have to put the css file in the same folder as the (to be generated) html one (or you can just have an horrible home page).

fanglingsu:
Oh, nice to see that this works.
Wouldn't it be easier to put the css into the html file? I think heredoc could make the code easier to read and would avoid the `\t` (that are only cosmetics for the html).

Artymort:
I think people (I ?) might want to change the style later, and hardcoding it in the script makes this painful... The installation process would be easier, but at a cost.

While heredoc would avoid all the ```echo -e```, it would also breaks the indentation of the script, which in my opinion makes it less readable.
And although the tabs are useless in a html file (as well as the new lines), I wouldn't say the ```\t```  makes the script any harder to read, but at least it produces a readable html file. But you could replace them by one or two spaces to avoid the current "compact" look of the strings.

fanglingsu:
> I think people (I ?) might want to change the style later, and hardcoding it in the script makes this painful... The installation process would be easier, but at a cost.

Hm, I thought the examples to provide some nice extra functionality that can be used like they are or that demonstrate how to tweak vimb to the users need. Right, an external css might be easier to change. But a singel file solution is easier to read and understand.

Followinfg the UNIX-Philosophy it would be better to separate the inotify and the spawning of vimb from the generation of the html into separate scripts or to let the user decide when to show the html. But hey, if this works for you and helps you to maintain you bookmarks, this is OK.

fanglingsu:
I don't understand JavaScript too, but I've adapted your script to work with JavaScript so there is no need for additional script.
```javascript
/**
 * Converts the bookmarkfile under $XDG_CONFIG_HOME/vimb/bookmark into nice html.
 * This script should be put into scripts.js file in vimb's config directory.
 * Open vimb's bookmark file in vimb file:///home/{user}/.config/vimb/bookmark and
 * run ':e! bookmarkFileToHtml();'
 * Or start the pretty view by autocmd in cofig file like
 * 'autocmd LoadFinished file:///*/bookmark e! bookmarkFileToHtml();'
 */
function bookmarkFileToHtml() {
    var parts=[], i, line, uri, lines = document.body.innerText.split(/\n/),
        html  = "<tr><th>Bookmark</th><th>Tag(s)</th></tr>",
        head  = '<head><style type="text/css">'
            + "body{background:#e0ddda}"
            + "table{margin:.5em auto;max-width:60em;border-collapse:collapse;color:#fff;}"
            + "th{text-align:left;background:#555;color:white;border-bottom:.3em solid #e0ddda;padding:14px 12px}"
            + "td{padding:8px 12px;border-bottom:1px solid #e0ddda;color:#555}"
            + "tr{background:#fff}"
            + "</style></head>";

    for (i = 0; i < lines.length; i++) {
        line  = lines[i];
        parts = line.split("\t");
        uri   = parts[0];
        title = parts[1]||uri;
        tags  = (parts[2]||"").split(' ').sort().join(', ');
        html  = html + "<tr><td><a href=\"" + uri + "\">" + title + "</a></td><td>" + tags + "</td></tr>";
    }

    html = "<html>" + head + "<body><table>" + html + "</table></body></html>";
    document.body.innerHTML = html;
}
```
Yes, this is only an example, and it's not very readable.

Artymort:
OK, the JS solution suits me well.
For a reason I don't know vimb seems to add a blank line at the end of the bookmark file, thus a strange empty line appears : if you publish it you might want to change the loop condition to 
```for (i = 0; i < lines.length - 1; i++) {```

And you definitely want to change the ```*``` in the path as it breaks the comment and the script !

**Comment by fanglingsu** [comment]
> For a reason I don't know vimb seems to add a blank line at the end of the
> bookmark file, thus a strange empty line appears

Or maybe we'd better check for any empty line.

> And you definitely want to change the * in the path as it breaks the comment
> and the script!

You're right, if written the script without the comment and added it only as I
posted it here.

**Comment by Artymort** [comment]
>>For a reason I don't know vimb seems to add a blank line at the end of the
>>bookmark file, thus a strange empty line appears

>Or maybe we'd better check for any empty line.

Or wonder why vimb adds an empty line at the end of a plain text file ? But this is out of the scope of this PR thread ;p

Anyway thanks for this !
I'm currently using it and it seems to be working just fine. I would love to find a way to open new tabs in background too, but I guess I should rather tweak i3 than vimb here. Or use the queue feature.

**Comment by fanglingsu** [comment]
> Or wonder why vimb adds an empty line at the end of a plain text file?

That is standard *NIX behaviour i think. Text files end up in a newline.

```
$ echo "foo" > test
$ wc -l test
1 test
$ od -c test
0000000   f   o   o  \n
0000004
```

The problem is the logic in JavaScript that splits the text at `\n` that
introduced the last empty line. That's a fault of the little script, no other
tool on *nix would see an additional or blank line on vimb's hisotry file.

> I'm currently using it and it seems to be working just fine. I would love to
> find a way to open new tabs in background too, but I guess I should rather
> tweak i3 than vimb here. Or use the queue feature.

Right, vimb does not know tabs. Each visible window is definitely a seperate
process that does not know anythings about possible other vimb instances. The
only connection between the processes exists in the data that are written to
the files in the config directory.
I use tabbed in case I want to have open multiple windows. This allows to open
new windows into background during compile time.

Artymort:
>Right, vimb does not know tabs. Each visible window is definitely a seperate
>process that does not know anythings about possible other vimb instances. The
>only connection between the processes exists in the data that are written to
>the files in the config directory.
>I use tabbed in case I want to have open multiple windows. This allows to open
>new windows into background during compile time.

Yes, I got that, I already have tabs using i3 in stack or tab mod by default for window of the class "Vimb" (I was using i3 anyway), the thing is that a new instance of vimb always grabs the focus, and in tabbed mode that mean being the only visible one (in his sub tree), while I like to open my tabs in background.

fanglingsu:
I've never used i3, so I can't help much here.

PR #195, opened 2015-03-19
Minor corrections in formfiller exemple

Artymort:
My dmenu does not know any ```-l7``` option, neither does the manual, I just added a space to make things work.

What's more I don't have any ```$XDG_*``` set by default, so I added an emptyness check and if the variable does not exist then the default value as recommanded [here](https://developer.gnome.org/basedir-spec/) is set.

Thanks for your work !


fanglingsu:
Thank you. I must confess that I haven't tested it after I changed it to use dmenu. I use a slightly different version that uses my global dmenu styling.

PR #200, opened 2015-04-20
enable input-mode on input.search element

semarie:
HTML5 define a new input type="search":
http://www.w3.org/TR/html-markup/input.search.html

declare this element to be editable, in order to switch in input-mode
when use it.

A possible webpage for test is http://doc.rust-lang.org/std/

### DIFF
--- a/src/dom.c
+++ b/src/dom.c
@@ -140,7 +140,8 @@ gboolean dom_is_editable(Element *element)
     if (!g_ascii_strcasecmp(tagname, "textarea")) {
         result = true;
     } else if (!g_ascii_strcasecmp(tagname, "input")
-        && (!*type || !g_ascii_strcasecmp(type, "text") || !g_ascii_strcasecmp(type, "password"))
+        && (!*type || !g_ascii_strcasecmp(type, "text") || !g_ascii_strcasecmp(type, "password")
+		|| !g_ascii_strcasecmp(type, "search"))
     ) {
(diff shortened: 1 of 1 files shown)


fanglingsu:
Thank you for this patch. Are there more fancy html5 input fields we should support?

semarie:
@fanglingsu the html5 specifications have severals new input type. According to http://www.w3.org/TR/html-markup/elements.html#elements, there are:
- input type=datetime – global date-and-time input control
- input type=datetime-local – local date-and-time input control
- input type=date – date input control
- input type=month – year-and-month input control
- input type=time – time input control
- input type=week – year-and-week input control
- input type=number – number input control
- input type=range – imprecise number-input control
- input type=email – e-mail address input control
- input type=url – URL input control
- input type=search – search field
- input type=tel – telephone-number-input field
- input type=color – color-well control

fanglingsu:
Wow, I think we should apply support for those of theme that require vimb to be in input mode. I think such things as color don't require vimb to be in input mode.

fanglingsu:
The decision which input types should switch vimb into input mode depends on the way webkit does handle the types. I found a nice test page with the different new input type fields http://robertnyman.com/html5/forms/input-types.html. Chromium handles most of them special, but vimb or better webkit does only handle `type="range"` and `type="number"` special. So I assume we have to support all new input types except of `range`.

PR #202, opened 2015-04-22
support html5 new input types

semarie:
mark new input types as editable, according to the way of webkit handle
these.

follow the proposition in #200


fanglingsu:
Tank you for helping to improve vimb. I've not much spare time at the moment and happy that you spent yout time for this project.

PR #217, opened 2015-06-20
Added hint-number-same-length option

yyt16384:
This option makes all hint numbers have the same length. With this option on there is no need to wait for `hint-timeout` or hit enter after entering the hint number, because it will not be ambiguous.


fanglingsu:
Thank you for this patch! This is a nice idea and really useful.

PR #219, opened 2015-06-25
Made build process more user-friendly

czarkoff:
Set CFLAGS that are not absolutely required with CFLAGS?=..., so that
user-defined CFLAGS take precedence.  Made VERBOSE option that would toggle
between silent (previous) and verbose mode of CC directive processing.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -9,6 +9,8 @@ EXAMPLEDIR ?= $(PREFIX)/share/$(PROJECT)/examples
 
 #----------------compile options---------------------
 
+VERBOSE ?= 0
+
 LIBS = libsoup-2.4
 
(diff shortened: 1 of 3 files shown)


czarkoff:
Nope, VERBOSE=0 would generate "silent rules" as autoconf calls it. Tested here.

czarkoff:
Sorry, misunderstood you.  Indeed, the blocks were in wrong order. FWIW all of those changes are for OpenBSD port of vimb.

fanglingsu:
Thank you for the patch. I'm no familar with the best practice of writing portable userfriendly Makefiles. So I'm happy about help doing things better.

Issue #224, opened 2015-07-16
Incorrect urlbar string when on a bad page


b80905:
If you open a nonexistent web address, the URL bar is not updated and the URL of
the previous page is displayed instead.

fanglingsu:
I can reproduce this behaviour. But what do you think should be shown in url bar instead, the uri that does could not be loaded, 'about:blank' or nothing?

fanglingsu:
Can you please check if the issue is fixed for you too?

b80905:
Not quite.
If you press f and then ESC on a nonexistent page, the URL of the
previous page is shown again.
If you open a nonexistent page and then another nonexistent one and
press CTRL+O, then the title of the previous page is displayed in the
title bar.

Of course the URL that couldn't be loaded should be shown.

fanglingsu:
You're right. I think the fix would take some time more, because this is webkit default behaviour. If a page could not be loaded the webview object contains the old uri and the old title. So we have to do some more internal logic for the title and current uri here.

fanglingsu:
So, another approach to fet rid of this issue. We'll see if this introduces some regression somewhere else.

fanglingsu:
This should be fixed now.

PR #226, opened 2015-07-17
Fix the title after going back to a local file

b80905:
If you open a local file, open a web page, and go back, the title
of the web page will still be shown. This is a fix.


fanglingsu:
Wow, you're a really good tester. I think this belongs to #224.

PR #227, opened 2015-07-18
Escape should deselect selected text

b80905:
If there's a text selection, pressing ESC will deselect it.
This does the same thing as #225, except that if there's no text
selection, the clipboard will not be cleared.


fanglingsu:
Ok, this works. I found a function that should do the same `webkit_web_view_delete_selection` but this does not delect the selected text even if the state is properly returned by `webkit_web_view_has_selection`. Your solution works well, but seems to be a little hacky.

PR #228, opened 2015-07-19
The input mode should be entered at the right time when a web page is being loaded.

b80905:
Suppose you are opening a web page on which the focus is automatically
given to a text box. WebKit gives it the focus immediately after the
text box appears on the screen. However, the input mode is entered
only after the page has been completely loaded. Until then, the normal
mode is active, so keys are processed by it first. This isn't the
expected behavior, since one shouldn't have to wait for the completion
of loading if one can already see the text box and wants to type
something in it.

### DIFF
--- a/src/dom.c
+++ b/src/dom.c
@@ -30,33 +30,27 @@ static gboolean editable_focus_cb(Element *element, Event *event);
 static Element *get_active_element(Document *doc);
 
 
-void dom_check_auto_insert(Document *doc)
+void dom_auto_insert_unless_strict_focus(Document *doc)
 {
-    Element *active;
-    HtmlElement *element;
+    Element *active = webkit_dom_html_document_get_active_element(WEBKIT_DOM_HTML_DOCUMENT(doc));
 
-    /* First check for current active element that becomes focused before we
-     * could add the evnet observers. */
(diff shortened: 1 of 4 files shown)


fanglingsu:
The intension is right, but the automatic input mode switching is a dificoult thing. There where a lot of issue in the past that where related to the automatic switching of mode on active form field or on leaving a form field.
1. It's not sure if the form field to focus is already visible in case the first visual layout event is dispatched.
2. The patch introduces a regression in the processing of `:set strict-focus=true`. The intension of this is to avoid the page to foce vimb into input mode. But with the patch vimb switches always to input mode if the page says so, independent from `strict-focus` setting. You can check this by the manual test page under _tests/manual/112-editable-focus.html_. I've not tested this, but you should test also _tests/manual/201-editable-focus-in-iframes.html_

I think I should write down some test cases to check vimb against. But I don't have much time, so this will take some weeks or more.

b80905:
This time I've taken the strict-focus variable into account. Both
of the tests that you mentioned were passed. What do you think?

fanglingsu:
OK, I've pulled you changes, but remove the revert and the revert of revert.
Do you think we can avoid the use of the `vb.state.done_loading_page` flag in the logic. At the time this is only used for the instert mode precessing and it's not really obvious what the flag means without this knowledge. If we could find a solution to avoid the flag by checking a flag on the events element or the webview, I'd prefer this.

PR #229, opened 2015-07-20
Stop focus from appearing to be on input box after closing external editor

b80905:
Open a web page with a text box. Use ";e" to open the text box in an external editor. Close the external editor. There will be a blinking cursor in the text box, even though it is not focused. This patch seems to fix that behaviour.

### DIFF
--- a/src/dom.c
+++ b/src/dom.c
@@ -78,6 +78,14 @@ void dom_clear_focus(WebKitWebView *view)
 }
 
 /**
+ * Give the focus to element.
+ */
+void dom_give_focus(Element *element)
+{
+    webkit_dom_element_focus(element);
+}
+
+/**
  * Find the first editable element and set the focus on it and enter input
  * mode.
  * Returns true if there was an editable element focused.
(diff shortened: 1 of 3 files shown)


b80905:
Why would anyone want that? If you have closed the external editor, it usually means that you are done editing the text box. Why give the focus to a text box that you are not going to change?

fanglingsu:
1. I would like to use some features of editing the field provided by the webpage - some validation might be triggered on chaning the field (might not work if the field isn't focuse at the time the content changes).
2. I'm undecided and want to open editor again - maybe I'm the only user who does this.
3. I'd like to use webkit feature to focus next formfield by `<Tab>` or `<S-Tab>`.

On the opposit I see no benefit of leaving the input mode, if all is done you'd normally submit the form by `<Enter>`, this works also in input mode.

b80905:
You've convinced me. Now the focus is given to the text box after the external editor is closed.

fanglingsu:
Tank you for spotting out the mode switching issues and for providing patches for them. Nice work!

PR #233, opened 2015-08-02
Do not blur the focused textbox after alt-tabbing.

b80905:
The focused textbox should still have the focus after the user switches to another window and then switches back to Vimb.

### DIFF
--- a/src/dom.c
+++ b/src/dom.c
@@ -281,9 +281,10 @@ static gboolean auto_insert(Element *element)
 
 static gboolean editable_blur_cb(Element *element, Event *event)
 {
-    if (vb.mode->id == 'i') {
+    if (vb.state.window_has_focus && vb.mode->id == 'i') {
         vb_enter('n');
     }
+
     return false;
 }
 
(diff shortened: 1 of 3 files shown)


fanglingsu:
Thank you for spotting the issue. But the fix looks a little bit like overkill. I'm not sure, but on a first view, it should be enough to use the `focus_in_event_cb` and return true, so that the event is not bubbled up. I don't know if this would break something else, but we should try this solution first.

b80905:
If focus_in_event_cb() returns true, then the cursor in the textbox will still be visible after the user switches to another window, since we do not let WebKit hide it in that case.

fanglingsu:
> If focus_in_event_cb() returns true, then the cursor in the textbox will still be visible after the user switches to another window, since we do not let WebKit hide it in that case.

Does the cursor in textbox on inactive window disturb you or someone else? Is there a reason to hide it on focus out except of cosmetics?

b80905:
I don't think it would be disturbing to anyone, and I don't know of any reasons other than the fact that  the expected behavior is hiding the cursor when the window doesn't have the focus.
It would be great if you could post your email address somewhere. I'd like to ask a few questions.

fanglingsu:
OK, lets do it right. I thought we can do ite with less code, but you convinced me to use the standard expected behaviour we know from webkit and other tools too.

PR #243, opened 2015-10-12
Fixed file locking using fcntl.

### LINKED ISSUE #235: Cookies file randomly reset
I can't pinpoint the trigger, but in my setup the cookies file seems to get randomly reset, such that I lose all my cookies and have to re-login to all sites if I open a new window (existing windows retain their session cookies and continue to work).

This was discussed in #154 but I have fixed the issues related to the values of `cookie-expire-time` and `cookie-timeout`, and this does not seem to be related AFAICT. I've reproduced the issue on two separate laptops (both running arch linux, both running vimb 2.10 (although slightly different commits)).

For now I'd just be happy to hear if anybody else has encountered the same issue, otherwise it is possibly due to my particular setup (both laptops have the same dotfiles, packages, etc. so almost identical). Any other clues to help debug this would be appreciated, I'm at a loss for how to figure out what's going on.

yyt16384:
This fixes #235.

### DIFF
--- a/src/cookiejar.c
+++ b/src/cookiejar.c
@@ -90,7 +90,7 @@ static void cookiejar_finalize(GObject *self)
 
 static void cookiejar_init(CookieJar *self)
 {
-    self->lock = open(vb.files[FILES_COOKIE], 0);
+    self->lock = open(vb.files[FILES_COOKIE], O_RDWR);
 }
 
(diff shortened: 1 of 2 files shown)


fanglingsu:
Thank you for the patch, nice work.

Issue #214, opened 2015-05-17
Input mode and command mode mix up after going back in history


ghost:
Under some conditions, vimb interprets attempts to edit web input fields as commands.

To reproduce:
1) Open duckduckgo.com or any search engine
2) Search for any term and click on any result
3) Go back to the search page pressing CTRL-O
4) Place cursor on the search engine's input field

Expected result:
Vimb should go in input mode and allow the user to type in another search.

Actual result:
Vimb allows the user to type in another search as long as the key is not bound to a vimb command. If the key is bound, vimb will execute the bound action. For example, attempting to type "r" will reload the page.

fanglingsu:
Tank you for this bug report. I can reproduce this, but may be it will take some weeks to fix this. There where a lot of input switching related issue in history of vimb and finxing one might generate a regression on some other page.
We had previously multiple parts in the code that where responsible to switch vimb int inputmode if a editable element was clicket or hinted. But the warious parts did their job different. Now there is only one place for the locig of switching it input mode. This is set as dom observer on that is added on the webviews `onload-event`.
I think the main problem is that the onload event isn't called in case vimb go back in history. There must be used some other event instead.

fanglingsu:
If I use a static page without fancy JavaScript with input element and leave this and go back sith `Ctrl-O` this switching to input mode on click works pretty fine.
I can't reproduce the issue if I disable page chache by `:set pagecache=0`, can you confirm this?

scrzbble:
It does not make a difference here whether I use java or not. I tried searching from https://duckduckgo.com/html and I still get the input/command bug.

However, I can confirm that disabling caching is a workaround.

shioyama:
I'm noticing this issue as well. Turning off caching for now but would be great to get a proper fix for it. (Vimb is great otherwise, thanks for all your work on this!)

shioyama:
@scrzbble #246 is merged now, if you have a chance could you try the latest vimb and confirm that it fixes the problem?

fanglingsu:
Seems to be fixed.

Issue #112, opened 2014-10-14
2.8: GLib-GObject-CRITICAL


ghost:
Started appearing periodically after updating to 2.8.

(vimb:49351): GLib-GObject-CRITICAL **: g_closure_unref: assertion 'closure->ref_count > 0' failed

fanglingsu:
I've compiled webkitgtk with debug symbols and got following backtrace.
    
    #0  0x00007ffff2ca3a80 in g_log () from /usr/lib/libglib-2.0.so.0
    #1  0x00007ffff700d7f8 in operator= (optr=0x0, this=0x7ffff7e49ed0) at ../webkitgtk-2.4.7/Source/WTF/wtf/gobject/GRefPtr.h:142
    #2  WebCore::GObjectEventListener::gobjectDestroyed (this=0x7ffff7e49ea8)
        at ../webkitgtk-2.4.7/Source/WebCore/bindings/gobject/GObjectEventListener.cpp:61
    #3  0x00007ffff3861ccf in weak_refs_notify () from /usr/lib/libgobject-2.0.so.0
    #4  0x00007ffff3862d65 in g_object_unref () from /usr/lib/libgobject-2.0.so.0
    #5  0x00007ffff700c9cc in WebKit::DOMObjectCache::clearByFrame (frame=<optimized out>)
        at ../webkitgtk-2.4.7/Source/WebCore/bindings/gobject/DOMObjectCache.cpp:109
    #6  0x00007ffff679f173 in WebCore::FrameLoader::commitProvisionalLoad (this=0x7ffff7e68c88)
        at ../webkitgtk-2.4.7/Source/WebCore/loader/FrameLoader.cpp:1813
    #7  0x00007ffff678699a in WebCore::DocumentLoader::commitLoad (this=0x7ffff7e7c900, 
        data=0x6c81d0 "<html>\n<body>\n<table><th align=\"left\">Name</th><th>Size</th><th>Date Modified</th>\n", length=8192)
        at ../webkitgtk-2.4.7/Source/WebCore/loader/DocumentLoader.cpp:765
    #8  0x00007ffff676be09 in WebCore::CachedRawResource::notifyClientsDataWasReceived (this=0x7ffff7ef5040, 
        data=0x6c81d0 "<html>\n<body>\n<table><th align=\"left\">Name</th><th>Size</th><th>Date Modified</th>\n", length=8192)
[...]

Issue #95, opened 2014-07-31
Some js pages exits input mode


ghost:
Some pages like translate.google.com exits input mode after you put any character in text field. Same goes for passthrough mode.

PR #246, opened 2015-10-22
Install focus/blur callbacks when load finished

shioyama:
Looking into #214 I found that the focus/blur callbacks were not being set when fetching from the cache because the `WEBKIT_LOAD_FIRST_VISUALLY_NON_EMPTY_LAYOUT` status is skipped, going directly to `WEBKIT_LOAD_FINISHED`, so added the call to `dom_install_focus_blur_callbacks` there as well.

Not very familiar with this code and not sure if it's okay to be adding listeners twice (in non-cache situations), but anyway this is a starting point.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -713,12 +713,12 @@ static void webview_download_progress_cb(WebKitWebView *view, GParamSpec *pspec)
 static void webview_load_status_cb(WebKitWebView *view, GParamSpec *pspec)
 {
     const char *uri;
+    WebKitWebFrame *frame = webkit_web_view_get_main_frame(view);
 
     switch (webkit_web_view_get_load_status(view)) {
         case WEBKIT_LOAD_PROVISIONAL:
 #ifdef FEATURE_AUTOCMD
             {
-                WebKitWebFrame *frame     = webkit_web_view_get_main_frame(view);
                 WebKitWebDataSource *src  = webkit_web_frame_get_provisional_data_source(frame);
                 WebKitNetworkRequest *req = webkit_web_data_source_get_initial_request(src);
                 uri = webkit_network_request_get_uri(req);
@@ -737,7 +737,6 @@ static void webview_load_status_cb(WebKitWebView *view, GParamSpec *pspec)
             autocmd_run(AU_LOAD_COMMITED, uri, NULL);
 #endif
             {
-                WebKitWebFrame *frame = webkit_web_view_get_main_frame(view);
                 JSContextRef ctx;
                 /* set the status */
                 if (g_str_has_prefix(uri, "https://")) {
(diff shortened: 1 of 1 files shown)


fanglingsu:
Wow, so easy! I had so many trouble in the last months with the input mode switching with some strange cornercases. But your fix seems not to make a regression for the cases that I can rember.
Google translate still works #95.
The manual test cases for the input switching still work.

Thank you for this nice patch!

fanglingsu:
In case of duckduckgo this fixes the issue when the page if fully loaded. sometime I run into cases where I go back and see the previous page the the input field but the page is still loading. Than I have to wait unti the page is fully loaded to be able to switch to input ode by mouse.
But is much better than being not able to get back into input mode with the mouse. Maybe we can find an event that is triggered right where we can apply the focus listeners.

shioyama:
Thanks for merging!

> In case of duckduckgo this fixes the issue when the page if fully loaded. sometime I run into cases where I go back and see the previous page the the input field but the page is still loading. 

Hmm if the page is cached I wonder why it would be loading.

>  Maybe we can find an event that is triggered right where we can apply the focus listeners.

Yes I have the feeling that this is still not quite right, there should be a place we can call it once for all situations.

fanglingsu:
> Hmm if the page is cached I wonder why it would be loading.

I assume that the cache contains only the html of the page, but that some scripts on the page and maybe some images are requested again. Maybe I've done something other than going back, but also a cache pages could spent some time loading something until the layout is loaded completely.

> Yes I have the feeling that this is still not quite right, there should be a place we can call it once for all situations.

In theory the right time to setup our listeners is the time when the dom is loaded. I've not tested it, but maybe we can observe the dom load in dom.c by something like
```c
WebKitDOMDOMWindow* domWindow = webkit_dom_document_get_default_view(doc);
webkit_dom_event_target_add_event_listener(
    WEBKIT_DOM_EVENT_TARGET(domWindow), "load", G_CALLBACK(...), false, NULL);
```
And do the things that are done in `dom_install_focus_blur_callbacks` in the callback. The dom load observer could be established on `WEBKIT_LOAD_FIRST_VISUALLY_NON_EMPTY_LAYOUT`.

fanglingsu:
Maybe the none working dom event listener on caches pages is related to #112. I've added a `g_object_ref(element);` into `dom_install_focus_blur_callbacks` and removed the additional event installation on load finished and this still works. By the way the `g_object_ref` fixed the #112 issue.

I know that this is a hack and I don't know where we should run the required `g_object_unref` to free the memory. But the `g_object_ref` seem to be a trick to make the code work like I've expected it since months so that the event observers are listening as long as the dom document exists somewehre in the browser.

Issue #138, opened 2014-11-01
Tabbed slow rendering?


dansbandit:
tried to run `tabbed -c vimb -e` from the documentation but it is not responsive at all. Tried opening a web page with tabbed and 2 tabs and without tabbed. I'm using the latest git branch for vimb and tabbed. I got some errors below but I don't know if they're related to the issue.

```
(vimb:5473): Gdk-WARNING **: vimb: Fatal IO error 0 (Success) on X server :0.

(vimb:5563): Gdk-ERROR **: The program 'vimb' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 7215 error_code 3 request_code 18 (core protocol) minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)

(vimb:5454): Gdk-ERROR **: The program 'vimb' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 3554 error_code 3 request_code 18 (core protocol) minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the GDK_SYNCHRONIZE environment
   variable to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
```

semarie:
Daniel,

I have switched from gtk2 to gtk3 for vimb (by compiling with GTK=3), and I have the same problem now. I will investigate this issue, and report elements.

For now:
 - the application seems inresponsive at loading, or when scrolling.
 - it occurs on a lot of pages (remote [http://] or local [file://]).
 - when hanging, if I switch vimb in background (the window is hidden), and go back in foreground, the rendering is actualised immediatly.
 - no errors in terminal nor Xorg log file.

It seems to be something like a buffering problem (display is not actualised).

fanglingsu:
[...]

PR #254, opened 2015-11-17
Small changes to build on OSX with macports

plredmond:
I built vimb on OSX Mavericks.

Steps to repro:

1. Install [macports](https://www.macports.org/install.php). Take note of the prefix (default is `/opt/local`).
1. Install vimb dependencies `sudo port install webkit-gtk3`. This takes hours.
1. Check out `vimb` https://github.com/fanglingsu/vimb/commit/c5b20658e84ae96c9945c1ca328813df8163aecc (last commit before switching to flock locking).
1. Apply the very small patch in this PR.
1. Build by telling `pkg-config` where to find the `vimb` dependencies:

    ```sh
    env PKG_CONFIG_PATH=${MACPORTS_PREFIX?}/lib/pkgconfig make
    ```

    If you have trouble with the library version conditional logic in `config.mk`, apply [this patch](https://gist.github.com/plredmond/c57bed437da15de6e4e5).
1. Feel vimb zen.

screen shot 2015-11-17 at 10 50 44

### DIFF
--- a/src/js2h.sh
+++ b/src/js2h.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-echo -n '#define HINTS_JS "'
+echo '#define HINTS_JS "' | tr -d '\n'
 cat $1 | \
     tr '\n\r\t' ' ' | \
     sed -e 's:/\*[^*]*\*/::g' \
--- a/src/main.h
+++ b/src/main.h
@@ -367,7 +367,7 @@ typedef struct {
     VbColor              status_bg[VB_STATUS_LAST];
     VbColor              status_fg[VB_STATUS_LAST];
     PangoFontDescription *status_font[VB_STATUS_LAST];
-} Style;
+} VbStyle;
 
 typedef struct {
     Gui             gui;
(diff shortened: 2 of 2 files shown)


plredmond:
A prerequisite for merging this PR might be fixing the conditional logic in `config.mk`. I'm not sure how to debug things in `make`, though.

fanglingsu:
Thank you for the patch. Looks good.

dan2:
Can confirm that this works in Yosemite as well.  (am posting this comment from vimb)

It does run very slowly, though.  Any thoughts on how to speed things up?

dan2:
Scratch that -- it isn't slow; the screen isn't updating.  I''m running vimb inside tabbed and if I switch from one tab and back to the previous, vimb has responded to my commands

plredmond:
I encountered the same issue. What versions of WebKit and gtk did you
compile with?

dan2:
gtk3 @3.20.3_0+x11
webkit-gtk3 @2.4.11_0+video

Installed both via macports

fanglingsu:
@dan2 I think that's the same issue as #138. But we don't know what's the reason for that. It seems to be some strange corner case between gtk3 and tabbed. You have two options here, compile against gtk2 or run without tabbed.

plredmond:
Despite my instructions to apply a patch in step 5 of the opening post for this PR, I think I worked around the "slow rendering" issue (when using `vimb` with `tabbed` as described in #138) by using `GTK2LIBS=gtk+-2.0 webkit-1.0`. I don't have access to that system anymore, and and cannot test this.

PR #255, opened 2015-11-26
Added closed-max-items option

yyt16384:
This option controls how many closed pages will be saved. Previously only 1 item is saved. It can also be set to 0 to disable saving closed pages.

Note that the behavior now is different from the previous one. Previously the `closed` file is not cleared when reopening the page, so it is possible to reopen the page several times. Now pages are only reopened once.


yyt16384:
It's updated now. I'm only moving pop logic to util, as the queue logic is already very simple (only a call to a util_* function).

fanglingsu:
Thank you for the patch. Pretty nice.

PR #274, opened 2016-03-03
Add support for per-domain user style sheets

tmathmeyer:
per-domain css files are kept in the XDG_CONFIG/vimb/styles directory.
They are named as [domain].css. For example, a custom style would be
named www.reddit.com.css. This user css file replaces the default
styles.css file, which is still loaded on sites with no user-specified
stylesheet.

This patch was born of an attempt to use the scripts.js file to load in css
per-domain, which basically can't work if that site uses the
document.onload and window.onload functions.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -771,6 +771,39 @@ static void webview_load_status_cb(WebKitWebView *view, GParamSpec *pspec)
                 js_eval_file(ctx, vb.files[FILES_SCRIPT]);
             }
 
+            {
+                const char *no_protocol = NULL;
+                if (g_str_has_prefix(uri, "https://")) {
+                    no_protocol = uri + 8; /* length of 'https://' */
+                } else if (g_str_has_prefix(uri, "http://")) {
+                    no_protocol = uri + 7; /* length of 'http://' */
+                }
+                if (no_protocol) {
+                    /* we need to set a character null, so duplicate, must free later */
+                    no_protocol = strdup(no_protocol);
+
+                    /* search for a / char. If we find it, set it null (to terminate string) */
+                    char *first_slash = g_strstr_len(no_protocol, -1, "/");
+                    if (first_slash) {
+                        *first_slash = (char)0;
+                    }
+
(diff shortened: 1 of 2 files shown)


fanglingsu:
I'm undecided. Well, it's nice to allow to setup the styles or the scripts file more fine grained. But I don't like the idea to use file called by the host for which they apply to.
In my opinion it would be more flexible to allow the user to configure the path to the scripts and style files during runtime. So it would be possible to use the auto command feature to setup them for different hosts or protocols. Ok, this would be a little bit harder to implement, but would allow to switch the files temporarily.
By the way, if you use https://developer.gnome.org/libsoup/stable/SoupURI.html functions you could reduce the code here `soup_uri_new()` and `soup_uri_get_host()` should work here, and the soup libs are already linked.

tmathmeyer:
Actually I agree; I think it would be interesting to try out, though I also feel that its a rather challenging decision as to what is the most usable. Also I'll check out the soup_uri_get_host.

yyt16384:
It may be better (and easier) to simply make the path a setting. Users can use autocmds to change it automatically.

So something like `set stylesheet-path=/path/to/style.css`, and call a script to change the setting in a `LoadCommited` autocmd.

fanglingsu:
@yyt16384 To setup the path to the stylesheet is a great idea. But I fear also that vimb becomes more and more complicated in the way the config files are processed. There is a default config file at the time and an option to use a custom config file. Some time ago a option to start vimb with a given profile was added that will create a directory within _$XDG_CONFIG_HOME/vimb/_ with it's own config file. And to allows to set the styles file (and the scripts file) path brings some more flexibility but also inconsistency in the way vimb is configured. I like the idea to switch file path's on runtime, but I hope we can find a solution that will cover also the current behaviour.
I've not thought about very conscientious, but it could be a choice to allow to set the path to all runtime files except of the config file during runtime. This would allow to set the style files path by autocommands and would also allow to emulate the profile feature (sterting vimb with custom config file which itself setup the paths to the other files).

PR #276, opened 2016-03-28
Use the beginning position of links for hinting

yyt16384:
Previously the top-left corner is used for creating hints. For links that span over several lines this is usually not the beginning of text. If there is another link at the beginning of this line, the two hints will be in the same place, making one of them hard to see.

I also changed the comparison logic in `isVisible`, because having 0 pixel visible should not be considered visible. It is included here because `getBoundingClientRect` will return (0,0,0,0) when `getClientRects` is empty, which was not caught in the `isVisible` check. This should also fix some hints appearing in top-left corner when no hintable elements are there.

### DIFF
--- a/src/hints.js
+++ b/src/hints.js
@@ -90,8 +90,8 @@ Object.freeze((function(){
                 }
                 var rect = e.getBoundingClientRect();
                 if (!rect ||
-                    rect.top > offsets.bottom || rect.bottom < offsets.top ||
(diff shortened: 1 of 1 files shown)


fanglingsu:
@yyt16384 Thank you for the patch, this works great. I've never noticed that there is a `e.getClientRects` method that does the trick. Pretty nice work.

PR #280, opened 2016-04-11
Adding additional CPPFLAGS to fix compilation errors on OpenBSD

fcambus:
Hello,

Since commit d546d7d1e39db8bac27cac250c3a40f854c1a9ef vimb uses flock instead of fcntl for file locking. This cause the following compilation error on OpenBSD :

```
cc hsts.o
hsts.c: In function 'save_entries':
hsts.c:445: warning: implicit declaration of function 'flock'
hsts.c:445: error: 'LOCK_EX' undeclared (first use in this function)
hsts.c:445: error: (Each undeclared identifier is reported only once
hsts.c:445: error: for each function it appears in.)
hsts.c:454: error: 'LOCK_UN' undeclared (first use in this function)
Makefile:47: recipe for target 'hsts.o' failed
gmake[1]: *** [hsts.o] Error 1
```

This diff fixes the issue.

Regards.


fanglingsu:
Hm, this does not work with gcc on my system.

    /usr/include/features.h:148:3: warning: #warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE" [-Wcpp]
    # warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"

It seems gcc want's to have `_DEFAULT_SOURCE` instead. Did the `CPPFLAGS += -_DEFAULT_SOURCE` work for openBSD too?

fcambus:
Unfortunately no, with _DEFAULT_SOURCE I'm getting the compilation errors again. Let's forget about this then, the OpenBSD vimb port has a patch for config.mk to tweak some paths, so I will just leave the CPPFLAGS change there.

Sorry for the noise.

iamleot:
Hello Frederic and Daniel!

Frederic Cambus writes:
> Unfortunately no, with _DEFAULT_SOURCE I'm getting the compilation
> errors again. Let's forget about this then, the OpenBSD vimb port
> has a patch for config.mk to tweak some paths, so I will just leave
> the CPPFLAGS change there.
>  
> Sorry for the noise.
AFAIK flock(2) is not portable[0]. Also in pkgsrc we've added a
conditional `CFLAGS` to add `-D_NETBSD_SOURCE` in order to access the
`LOCK_*` macros in `<fcntl.h>`.

I think that apart doing it in the port/pkgsrc Makefiles (probably
the easiest and less intrusive way, but still pretty ok IMHO) the
respective `-D` (e.g. `-D_BSD_SOURCE` for OpenBSD, `-D_NETBSD_SOURCE`
for NetBSD, etc. if also other OSes are affected) can be added
conditionally depending by the OS. Maybe we can parse the `uname -s`
output and handle the several cases?

If you'd like this approach please let me know and I can give a look at
`config.mk` to implement that.


Ciao,
L.

[0]: there is a nice article regarding it by Jonathan Perkin if you
 are more interested:
 <https://www.perkin.org.uk/posts/solaris-portability-flock.html>

fanglingsu:
@iamleot I'm not sure about the approach, but help is pretty welcome. I hope that there is a solution for right file locking that is more portable than the `flock` and special precompiler flags for each system.

PR #293, opened 2016-09-15
Feature: set initial default zoom via config

rti:
This feature adds a new configuration value:

`default-zoom` This value represents the initial Full-Content Zoom level in percent. Default is 100.

So you are able to scale content up or down by default, without pressing `zI`/`zO`.

I implemented this feature for all those who:
 - need a bit more size to please their eyes
 - have a hard time configuring their high resolution monitor setup properly :)

Hope this might help someone.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1295,6 +1295,9 @@ The value `0' convert all cookies as session-only cookies
 Any other value enforce the expire-time (the expire-time value will be the
 lower between server-side request and time defined with `cookie-expire-time').
 .TP
+.B default-zoom (int)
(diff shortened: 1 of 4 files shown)


fanglingsu:
Thank you for your work. This works fine. I don't have access to a high-DPI screen to check if we could replace the high-DPI stuff by the default zoom.

rti:
Thank you for vimb!

Well, I think on a properly configured high DPI setup, the existing automatic approach querying gdk is more elegant. It just did not work on my machine, but this should be due to my config... `xrandr` and `xdyinfo` are reporting different screen sizes. Wired stuff going on...

PR #300, opened 2016-10-01
Fix default previouspattern

Nicop06:
This fixes a type in the previous pattern regex, which didn't work


fanglingsu:
Thank you for your patch.

PR #301, opened 2016-10-01
Fix default previouspattern in doc

Nicop06:


fanglingsu:
I've merge the patch, but there where to many backslashes that where only needed in the settings to escape backslashes and pipes.

Nicop06:
Thanks for the merge!

Sorry, I just copy pasted the config. I should have checked the output with jekyll.

fanglingsu:
No problem. Thank you for your work.

PR #304, opened 2016-10-12
Feature: webkit2 downloads

rti:
This implements some convenience wrapping around webkit's file downloading functionality

 * Configure a download path via settings (`~` and env vars supported)
 * Name unnamed download files
 * Ensure unique filenames by inserting a numerical suffix at a sane position
 * Track downloads in client state
 * Do not quit if downloads are running
 * Report download success and failures via status bar

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -49,7 +49,14 @@ static void input_print(Client *c, gboolean force, MessageType type,
         gboolean hide, const char *message);
 static void marks_clear(Client *c);
 static void mode_free(Mode *mode);
+static void on_webctx_download_started(WebKitWebContext *webctx,
+        WebKitDownload *download, Client *c);
 static void on_webctx_init_web_extension(WebKitWebContext *webctx, gpointer data);
+static gboolean on_webdownload_decide_destination(WebKitDownload *download,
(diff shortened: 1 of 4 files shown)


fanglingsu:
Thank you for this patch.

src/main.c
```diff
@@ -744,6 +766,168 @@ static void on_webctx_init_web_extension(WebKitWebContext *webctx, gpointer data
...
+            suffix_pos = extension_dot - filename;
+        } else {
+            suffix_pos = strlen(filename);
+        }
+
+        /* construct a new complete download filepath and add the suffix before
+         * the filename extension, keep incrementing the suffix value as long
+         * as the filepath exists, stop on first unused filename. */
+        do {
+            g_string_printf(filepath, "%s%c%.*s_%i%s",
```
fanglingsu: In my opinion, the filepath have to be freed during looping by `g_free()`. I think, there must be an `g_free(filepath);` before the `g_string_printf`.

fanglingsu:
You are right, i was a little tired as i saw a memory leak there.

rti:
Always good to double check.

PR #307, opened 2016-10-18
Fix/webkit2 config.h auto update

rti:
This patch adds a `make` rule to automatically update `config.h` if `config.def.h` changed. As user changes in `config.h` will be overwritten this way, a backup of the previous `config.h` is created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,11 @@ options:
 	@echo "LDFLAGS = $(LDFLAGS)"
 	@echo "CC      = $(CC)"
 
-vimb: $(SUBDIRS:%=%.subdir-all)
+$(SRCDIR)/config.h: $(SRCDIR)/config.def.h
+	@cp $@ $@.`date +%Y%m%d%H%M%S`
+	@cp $< $@
+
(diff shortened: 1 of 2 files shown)


fanglingsu:
I personally find it easy to put changes lines from config.def.h into my config.h and this is also the way the suckless tools are compiled. I don't see a benefit from writing config.def.h into config.h and backupf the config.h each time. I assume you intend to avoid cases where a user might have missed to check for essential changes in config.def.h and files a bug report. But the user has to check the config even with your patch, else the customized configuration will be left and this is not as obvious visible to the user like a compile error.

rti:
I see. It's also a very valid point. 

I mainly looked at the trivial `git clone $URL; make` case. Did not like this one failing on the user right away. 

Anyway. It's just a matter of taste.

fanglingsu:
This is only an issue in case you make the `git clone $URL; make` in an none
empty directory, else the config.h is generated on the first run of make.

rti:
Really? Oh, didn't know that. I am fine anyway then :)

PR #308, opened 2016-10-18
Feature/webkit2 gui style settings

rti:
This patch implements CSS GUI style settings on the `webkit2` branch.

Via `config.h`, two modes can be selected:
 * **vimb2 compatibility** (current default) to still allow vimb2 config files to work properly, very handy on machines where vimb2 is in production and vimb3 is in development
 * **vimb3** with less settings and more freedom, where each setting holds the complete css string for a GUI element in a certain state

Later, the vimb2 compatibility mode could be removed - for vimb3 release for example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ project directory.
   - [ ] try to use the webkit2 feature of running multiple pages with related
         view instance `webkit_web_view_new_with_related_view`
   - [ ] allow setting of different scopes, global and instance (new feature)
-  - [ ] remove the settings related to the gui like `status-color-bg` this was
+  - [x] remove the settings related to the gui like `status-color-bg` this was
         only a hack and is not recommended for new gtk3 applications. the
         color and font settings should be setup by css instead.
   - [ ] webkit2 does not provide the view mode `source` so maybe this is going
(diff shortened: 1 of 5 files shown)


fanglingsu:
Nice work! 
Hope we can remove the compat part soon. I'm not really sure if it where a good idea to make the gui style configurable on runtime. I never touch these settings after compiling, but there are cases where a color change might be a nice indicator for special configured browser window.
I've removed the duplicate trailing `;` in the css rules - in fact there is no trailing `;` needed. And I also added the two constants to set the normal and emphasized gui font from within config.h.

rti:
Thanks a lot for the merge and for the adjustments. 

As I see vimb, it is nicely designed to be straight forward, minimalistic, still easy to use (for vimified users). Having a runtime config let's the user play around and customize without having a compiler and gtk3 headers installed. I find this very important. The suckless approach excludes all the geeky users not fancying to need a compiler all the time. Also, the suckless design is incompatible with basically every package manager. Vimb perfectly fills a gap here. 

"Yes, yes, but colors at runtime..." you say? Well, I personally have window manager, terminal, launcher, and now also browser ;) consistently styled using https://github.com/chriskempson/base16 . For me, colors are an essential part of the user configuration.

Issue #247, opened 2015-10-22
Don't translate keyevent into char


fanglingsu:
At the time vimb uses a similar approach to process the keybinds like vim. The keyevents are translated in a array of unsigned chars. This has some disadwantages:
- Some keys can't be distinguished for example `<C-M>` and `<CR>` that are converted into the same char. Or `<C-S-T>` and `<C-T>` which are the same.
- The conversion is a little mess at the time and not really obvious to new developers that don't know the internals if vim

So we should follow [the suggestion for vim](http://vim.1045645.n5.nabble.com/Keyboard-input-handling-td1211322.html) also for vimb and use the keyval and state of the key event's internally.

MichaelMackus:
:+1: 

I've added a PR with rudimentary support for this "queue" concept. Current goal of the PR is just to fix #258, but could eventually be moved to be the central processing logic for keybinds.

fanglingsu:
I'll keep the PR open because at the time this is not ready to be used. So you can push further commits to you branch. I'm not sure about the restriction to input mode. If this works, I would try to do it in all modes.

I'm not sure if this approach is the right way, because this works only in case we get GdkEvents. If the events came from socket or from another keybinding this would not work (`nmap foo giInsert This Text` - `gi` set the cursor into first form field and enters input mode and the "Insert This Text" should be written into the form field). I know that is an other issue, but I hope that we find a solution that solved both cases without adding to much bloated code. So I see your branch as an proof of concept to check if this works and to experiment with queued key event (not inevitably GdkEvents).
Thank you for your time to help making vimb more useable.

PR #311, opened 2016-11-01
[WIP] GdkEventKey queue

### LINKED ISSUE #258: Imap which “give up” when not matching (like vim)
I configured vimb and vim to exit insert mode when I press “aa” like this
`:imap aa <Esc>`
But unlike vim, vimb don’t input any a anymore…

vim
« I’ve got a computer » -> « I’ve got a computer »

vimb
« I’ve got a computer » -> « I’ve got  computer » # with 2 space and no a

Not been able to type any a is not very fancy…

MichaelMackus:
Goal is to fix #258 

Related: #247

### DIFF
--- a/src/events.c
+++ b/src/events.c
@@ -1,0 +1,66 @@
+#include "events.h"
+
+/* this is only to queue GDK key events, in order to later send them if the map didn't match */
+static struct {
+    GdkEventKey **queue;                /* queue holding submitted events */
+    int            qlen;                /* pointer to last char in queue */
+    bool           processing;          /* whether or not events are processing */
+} events;
+
+/**
+ * Append an event into the queue.
+ */
+void queue_event(GdkEventKey* e)
+{
+    if (vb.mode->id != 'i') {
+        /* events are only needed for input mode */
+        return;
+    }
+
+    if (events.queue == NULL) {
+        events.queue = malloc(0);
+    }
+
+    events.queue = realloc(events.queue, (events.qlen + 1) * sizeof *events.queue);
+    events.queue[events.qlen] = e;
+    events.qlen ++;
+}
+
+/**
+ * Process events in the queue, sending the key events to GDK.
+ */
+void process_events()
+{
+    if (vb.mode->id != 'i') {
+        /* events are only needed for input mode */
+        return;
+    }
+
+    events.processing = true; /* signal not to map our events */
+
+    for (int i = 0; i < events.qlen; ++i)
+    {
+        // TODO send to gdk
+    }
+
+    events.qlen = 0;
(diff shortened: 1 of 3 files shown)


src/map.c
```diff
@@ -131,6 +132,11 @@ void map_cleanup(void)
  */
 gboolean map_keypress(GtkWidget *widget, GdkEventKey* event, gpointer data)
 {
+    if (is_processing_events()) {
+        // events are processing, pass all keys unmodified
+        return false;
```
fanglingsu: I think we have to queue this event too, else we could end up with key events put into wrong order - current one followed by some of the enqueued events.

src/map.c
```diff
@@ -131,6 +132,11 @@ void map_cleanup(void)
  */
 gboolean map_keypress(GtkWidget *widget, GdkEventKey* event, gpointer data)
 {
+    if (is_processing_events()) {
+        // events are processing, pass all keys unmodified
```
fanglingsu: Can you please use the `/* */` ansi C comments instead of the `//`.

src/map.c
MichaelMackus: This is only for when we send the events to GDK/GTK - see process_events() function

src/events.c
```diff
@@ -0,0 +1,103 @@
...
+    int            qlen;                /* pointer to last char in queue */
+    bool           processing;          /* whether or not events are processing */
+} events;
+
+/**
+ * Append an event into the queue.
+ */
+void queue_event(GdkEventKey* e)
+{
+    if (vb.mode->id != 'i') {
```
MichaelMackus: The reason I'm casing out input mode here is because command mode (and possibly other modes) seem to be respecting incorrect maps fine. The initial goal of this PR is just to fix #258. Eventually this will be available for all input modes.

src/events.c
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    free(events.queue[events.qlen - 1]);
+    events.qlen --;
+}
+
+/**
+ * Process events in the queue, sending the key events to GDK.
+ */
+void process_events(bool is_timeout)
```
MichaelMackus: Added bool here since we need to case out the popping of events. Goal is to remove this in a future commit.

MichaelMackus:
@fanglingsu from my testing, this is already fixing #258 - if I have "imap jk <Esc>" and I input "all work and no play makes jack a dull boy" I get that exact sentance back. When I hit "jk" it escapes out of the input field. Will now work on cleaning up the code.

fanglingsu:
I've adapted you code to use the `glib` memory allocation functions and the GSList.

PR #312, opened 2016-11-03
Allow conditional compilation of xembed

pks-t:
In order to make use of the xembed feature, we require
functionality native to the X11 windowing environment in order to
get a window's X identifier. Obviously, this requires us to
compile with a Gtk+ library which has support for X, restricting
us from working on Wayland-only platforms.

The issue can be fixed by making the Xembed feature optional. As
it is the only feature requiring platform-specific features, this
neatly allows us to build for a Wayland-only platform when XEMBED
is disabled.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -845,7 +845,7 @@ feature and started with `--socket' option.
 .TP
 .B VIMB_XID
 Holds the X-Window id of the Vimb window or of the embedding window if Vimb is
-started with the -e option.
+compiled with XEMBED and started with the -e option.
 .PD
 .PP
 Example:
--- a/src/config.def.h
+++ b/src/config.def.h
@@ -28,6 +28,8 @@
 #define FEATURE_SEARCH_HIGHLIGHT
 /* disable scrollbars */
 #define FEATURE_NO_SCROLLBARS
+/* disable X window embedding */
+/* #define FEATURE_NO_XEMBED */
 /* show page title in url completions */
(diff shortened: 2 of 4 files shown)


fanglingsu:
Thank you for the patch. I think the `#include <gdk/gdkx.h>` in main.c should be kept in case XEMBED is used, else the `GDK_WINDOW_XID()` is not known.
I'm not sure about this change because the XEMBED is an essential Feature in vimb and I assume there are many people using it. But if we insert the `FEATURE_XEMBED` will cause many users to not have this feature, because their config.h is not updated with the config.def.h. I know that this is strange assymmetric, but I thinks it would be better to use seomthing like `FEATURE_DISABLE_XEMBED`. This would not change anything to those users who alredy use vimb. Or even better, would be to detect on compile time if the xembed can be compiled in. If this is possible I'd prefer this automatic detection to always compile with xembed if this is available.

pks-t:
We already include `main.h`, through which we transitively include `gdk/gdkx.h`.
Regarding opt-in/opt-out/autodetection of `FEATURE_XEMBED`, I don't particularly care here. I'll update the PR later to change how we determine if we should compile with XEMBED or not.

pks-t:
Sorry it took so long to come up with this patch. The logic is now inverted. I called it `FEATURE_NO_XEMBED`, relating it to `FEATURE_NO_SCROLLBAR` as precedence.

fanglingsu:
Thank you for your work.

PR #313, opened 2016-11-12
fix path expansion to accept valid POSIX usernames

manzur:
To be portable across systems conforming to POSIX.1-2008, the
username is composed of characters from the portable filename character
set([A-Za-z._-])[1][2].

The <hyphen-minus> character should not be used as the first
character of a portable user name.

Links:
[1] Username format:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_435

[2] Portable filename character set:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

### DIFF
--- a/src/ascii.h
+++ b/src/ascii.h
@@ -60,16 +60,17 @@ static const unsigned char chartable[256] = {
 #undef C
 #undef SC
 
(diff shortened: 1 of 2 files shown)


fanglingsu:
Thank you for the patch. And sorry for the late response.

PR #317, opened 2016-12-26
Update formiller

rbialek85:
name='user' 

...as it is suggested, simply doesn't work, instead it must be:

name='login'

...as it is (correctly and already) in the file "scripts.js".

(BTW: I'm new to GitHub. This is my first pull request ever. I hope everything's fine, haha)


fanglingsu:
@rbialek85 This is only an example, the `["input[name='user']` is only a hint how to setup the element selector to select a form field. The 'user'-part might also be something else.

PR #318, opened 2016-12-28
Update README.md

mixmix:
There are a couple of other links that also need changing - I think those starting with https://fanglingsu.github.io/vimb/keybindings.html are legacy, but I'm not sure where some of them are meant to point to now


fanglingsu:
@mixmix Thank you for the patch. I've remove those old links because there is no good section to point them to.

mixmix:
thank you for vimb, it makes my heart sing.
a friend introduced me to it via a succinct blogpost http://blog.eight45.net/2016/12/26/vimb.html

I went looking for a way to block ads (it was the only major thing missing for me) and am trying out a hostfile solution which looks quite good: 
https://github.com/StevenBlack/hosts

PR #321, opened 2017-02-17
Webkit2 font size improvements

rti:
This changes aim to make the default font size settings look prettier and better match common settings know from vimb2 and other browsers.

1. The GTK GUI font size is now set to 10pt instead of 10px, so the GUI looks more similar to vimb2
2. The WebKIT2 default font and default monospace font sizes are now set to 16 and 13 respectively (the WebKIT2 defaults [1] [2])

[1] https://lazka.github.io/pgi-docs/WebKit2-4.0/classes/Settings.html#WebKit2.Settings.props.default_font_size
[2] https://lazka.github.io/pgi-docs/WebKit2-4.0/classes/Settings.html#WebKit2.Settings.props.default_monospace_font_size

### DIFF
--- a/src/config.def.h
+++ b/src/config.def.h
@@ -47,6 +47,6 @@
 /* default font size for fonts in webview */
 #define SETTING_DEFAULT_FONT_SIZE             16
 #define SETTING_DEFAULT_MONOSPACE_FONT_SIZE   13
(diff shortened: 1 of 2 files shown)


fanglingsu:
Thank you for the patch.

PR #322, opened 2017-02-20
webkit2: ex shortcut support

rti:
Adds support for the ```shortcut-add```, ```shortcut-remove``` and ```shortcut-default``` commands to ex.

### DIFF
--- a/src/ex.c
+++ b/src/ex.c
@@ -967,14 +967,10 @@ static VbCmdResult ex_handlers(Client *c, const ExArg *arg)
 
 static VbCmdResult ex_shortcut(Client *c, const ExArg *arg)
 {
-    char *key, *uri;
-    gboolean res = CMD_ERROR;
+    gchar *uri;
+    gboolean success = false;
 
     if (!c) {
-        return CMD_ERROR;
-    }
-
-    if (!arg || !arg->name) {
         return CMD_ERROR;
     }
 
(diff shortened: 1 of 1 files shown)


src/ex.c
```diff
@@ -966,8 +967,38 @@ static VbCmdResult ex_handlers(Client *c, const ExArg *arg)
...
+
+    if (!arg || !arg->name) {
+        return CMD_ERROR;
+    }
+
+    if (!arg->rhs || !arg->rhs->str || !*arg->rhs->str) {
+        return CMD_ERROR;
+    }
+
+    if (strstr(arg->name, "shortcut-add")) {
```
fanglingsu: Each `arg` has the `code` property that can be checked against `EX_SCA`, `EX_SCD` or `EX_SCR` so there is no need to do a string comparison here. I think you can use this function from the master branch except of the new client parameter and the returend values. I'd also prefer not to strip key and uri before adding to the shortcuts until we know that this is required. If there is a need for stripping, we should do this in `shortcut_add` so that possible other callers do not have to concern about stripping as well.

src/ex.c
```diff
@@ -966,8 +967,38 @@ static VbCmdResult ex_handlers(Client *c, const ExArg *arg)
...
-    /* TODO no implemented yet */
-    return CMD_SUCCESS;
+    char *key, *uri;
+    gboolean res = CMD_ERROR;
+
+    if (!c) {
+        return CMD_ERROR;
+    }
+
+    if (!arg || !arg->name) {
```
fanglingsu: I don't think this is required, there is no other function in which is checked if the `arg` is given and has a name. These function are only called by `ex_run_string` so we can add these checks there in case they will be required in future.

src/ex.c
rti: Thanks a lot for the hints, I learned a lot. I will push another update to align the implementation to the master branch's version.

Though, I would like to keep the stripping where it is, here is why:

1. We support ex commands such as ```set a=1``` and ```set     a = 1```, and they do exaclty the same (https://github.com/fanglingsu/vimb/blob/webkit2/src/ex.c#L910). I wanted to have the same functionality for ```shortcut-add```.
2. Allowing spaces after shortcut keys does not work as ```shortcut_lookup()``` splits key and value on the first space (https://github.com/fanglingsu/vimb/blob/master/src/shortcut.c#L197). Also tailing spaces on keys don't make any sense to me.
3. Moving ```g_strstrip()``` to ```shortcut.c``` would require changing the shortcut API as key and value paramters are ```const char*``` over there.

src/ex.c
rti: Thanks for the comment. Will be addressed with the next commit.

fanglingsu:
Thank you for your work! I hope there would some more developers help to get the webkit vimb working.

PR #325, opened 2017-02-27
Webkit2: ctrl+p and ctrl+n for ex history movement

rti:
Vim supports ```CTRL+p``` and ```CTRL+n``` for ex command history movement (and so do bash and zsh). 

Find attached the patch to add it to vimb as well.


fanglingsu:
Right I missed that. Never used this in vim.

Issue #122, opened 2014-10-21
Add :source command


fanglingsu:
Add the `:source` command like in vim to allow to execute ex commands from a file given file. It' would be nice to have completion for file path if run from vimb's inputbox.

Phenitei:
This issue was fixed by commits ffd7d6a57d18ea9d193ac20613cb5b9dfa973039 ffd7d6a57d18ea9d193ac20613cb5b9dfa973039 and ffd7d6a57d18ea9d193ac20613cb5b9dfa973039 whose additions to the source code were deleted during the great webkit2 purge. The reason for the deletion of the source command is not documented anywhere, is it voluntary or an omission? If it is an omission, I have a branch where I re-added the command and can open a pull request, if it is voluntary then this issue should probably be closed?

fanglingsu:
@Phenitei I've purged a lot from vimb only to reach a state where that was worth to be pushed - a state where vimb compiles. The intension is to migrate as many existing features to the webkit2 branch as possible. If you have some spare time please feel free to open a pull request.

Phenitei:
Opened #326 .

PR #326, opened 2017-03-01
Fix Issue 122

Phenitei:
As discussed in issue #122 I re-added the necessary code to use the :source command. Everything compiles and works fine.


fanglingsu:
Thank you to the pull request.

Issue #284, opened 2016-04-24
Shortcuts without arguments


aksr:
Currently, shortcuts without arguments don't do anything useful.
I doubt google search for `g` or `i` or `s` etc. is useful for anyone.
For example, _conkeror_ has `$alternative` for shortcuts without arguments, e.g:

```
 define_webjump("lwn.net", "http://lwn.net/Search/DoSearch?words=%s", $alternative="http://lwn.net");
```

Somethink similar would be nice to have.
What do you think?
Is there a way to implement this?

PS: $[0-9] is a really useful feature ;)

fanglingsu:
> I doubt google search for g or i or s etc. is useful for anyone.

You are right, to use a shortcut without arguments makes no sense. But I don't see the need to add code to do a fallback to something else. The user decides what vimb should do, or in this case what the shortcuts work. So I assume that the users know what they are typing in the statusbar. When I understand the webjump right, the path part of the url is stripped away, if the webjump is called without arguments (or the given alternative is taken). But in this case is nothing more than opening a bookmark to the alternative page.

aksr:
> You are right, to use a shortcut without arguments makes no sense. 

Of course, but it *could* make sense, it *could* be useful.
For example, w (for wikipedia):
`w` (without arguments) would go the wikipedia main page, e.g:

    https://en.wikipedia.org/wiki/Main_Page

Yet, `w Albert Einstein` searches wikipedia, e.g:

    http://en.wikipedia.org/wiki/Special:Search?search=Albert+Einstein

There are plenty examples like this (*imdb* news and search for title/actor etc.).
Besides Conkeror, Firefox has the same possibility using `OpenSearch`.

aksr:
Pity.

fanglingsu:
@aksr Can you have a look at the pull request #329 from @rti? Would this approach help you?

aksr:
@fanglingsu: **Well, it's better then nothing.** (While very few would search for *g* or *w* or *whatever*.)
Still, having/using only one shortcut (e.g. `w`) is easier and more elegant.
However, if it *very much* complicates the code to have something like this, then, I guess, it's not worth it.

Issue #181, opened 2015-02-17
implement suggestengines


257:
...vimperator style:
https://github.com/vimperator/vimperator-labs/search?utf8=%E2%9C%93&q=suggestengines&type=Code

fanglingsu:
I don't know what you have in mind, the search url you provided shown only a two matches in JavaScript that make me believe that there is one fork of vimperator-labs that llos to set a property called `suggestengines`.
What should be suggested, where and why? At the moment an can only guess that this might be related to #180.

foo-bar-baz:
On Feb 17  14:26 -0800, Daniel Carl wrote:
> I don't know what you have in mind, the search url you provided shown
> only a two matches in JavaScript that make me believe that there is one
> fork of vimperator-labs that llos to set a property called
> `suggestengines`.  What should be suggested, where and why? At the moment
> an can only guess that this might be related to #180.

it is indeed related to #180 but this is for the search suggestion only. i
asked people on #vimperator channel earlier and someone suggested that it
uses firefox's built-in suggestion engine which seems to be a module that
talks to google's autosuggestion api.

fanglingsu:
Ah you mean that you are not sure what to search for and start typing to near down the list of older search queries from other users to search for them.

At my current point of view this is to much overhead and a lot of work. And I don't think this will be added to vimb. Besides this I don't like the concept that a search engine suggests the user what to search for. If I'm interseted in something, I'll search for that and don't want to be persuaded to search for something other albeit something similar.

foo-bar-baz:
On Feb 18  14:30 -0800, Daniel Carl wrote:
> Ah you mean that you are not sure what to search for and start typing to
> near down the list of older search queries from other users to search for
> them.

no, that would be stupid.

> At my current point of view this is to much overhead and a lot of work.
> And I don't think this will be added to vimb. Besides this I don't like
> the concept that a search engine suggests the user what to search for. If
> I'm interseted in something, I'll search for that and don't want to be
> persuaded to search for something other albeit something similar.

i give you an example. i'm not a native English speaker. and learning the
language i discovered some articles are bitch to memorise in terms of where
they're used. for instance, would you say "i will see _at_ the party" or "i
will see you _on_ the party". looking at what others say in this case would
be helpful, me thinks. or if i have no notion of how Engligh speaker would
spell a Russian name like Tarkofsky (the probably spell Tarkovsky), well i
would like to type Tarko and see what i get. i can give you endless
[...]

PR #329, opened 2017-03-02
Allow shortcuts without parameters

rti:
This change provides a simple way to allow calling shortcuts without parameters.

I see several positive effects:

### Navigate to your favourite pages via shortcuts
Without necessarily going to search there but check the news feed or what so ever.

### Compromise for #284 
You could now do:
```
shortcut-add w=https://en.wikipedia.org/wiki/Main_Page
shortcut-add ws=http://en.wikipedia.org/wiki/Special:Search?search=$0
```
So you would have to define two shortcuts, but basically it allows the workflow @aksr requested.

### Compromise for #181

Say you set a duckduckgo short via:

```
shortcut-add du=https://duckduckgo.com/?q=$0
```

You can now ```:open du some search string``` as usual. But you can also just ```:open du``` which will bring you to the duckduckgo front page featuring search suggestions. Several search engines I tested send you to the search front page if the url query was empty. So it's a way to rapidly access search suggestions without vimb implementing it :) Maybe it helps the workflow of @257

What do you think?

### DIFF
--- a/src/shortcut.c
+++ b/src/shortcut.c
@@ -202,6 +202,8 @@ static const char *shortcut_lookup(Client *c, const char *string, const char **q
             *query = p + 1;
         }
         g_free(key);
+    } else {
+        uri = g_hash_table_lookup(c->shortcut.table, string);
     }
 
     if (!uri && c->shortcut.fallback


src/main.c
```diff
@@ -252,8 +252,8 @@ gboolean vb_load_uri(Client *c, const Arg *arg)
         rp  = realpath(path, NULL);
         uri = g_strconcat("file://", rp, NULL);
         free(rp);
-    } else if (strchr(path, ' ') || !strchr(path, '.')) {
-        /* use a shortcut if path contains spaces or no dot */
+    } else if (!strchr(path, '.')) {
```
fanglingsu: Is this really needed for the parameter free shortcuts? I've added the `strchr(path, ' ')` to allow to search for something containing dots like "document.activeElement". This would not work without the check for space.

rti: You are totally right, my bad. I misinterpreted the condition. I will update the commit.

rti:
Updated, not removing the check for space anymore.

fanglingsu:
Tank you!

aksr:
Could be this also merged into master branch?

fanglingsu:
@aksr I've applied the changes to the master branch too.

aksr:
Thank you.

PR #330, opened 2017-03-08
Fix for input box prompt color after error

rti:
When an error message is displayed by the input box, the next ```:open``` and ```:tabopen``` commands entered using ```o``` and ```t``` keys are rendered using input error color, not input normal color.

## Steps to reproduce
- run a non existing command, e.g. ```:asdf```
- input box will print ```Unknown command: asdf``` using input error color
- type ```t``` or ```o``` to open an URL

## Expected behaviour
- input box prints using input normal color

## Actual behaviour
- input box prints using input error color


fanglingsu:
Nice work, never have noticed this behaviour.

PR #336, opened 2017-04-08
Incsearch search while typing

rti:
This patch implements incremental search, or "search while typing", as known from vim as ```incsearch```. Hence, there is a new boolean setting called ```incsearch```, off by default.


fanglingsu:
Pretty nice!

PR #339, opened 2017-04-10
config.mk simplification

rti:
This patch removes the distinction between ```CFLAGS``` and ```CXXFLAGS``` from ```config.mk```. I think it is simpler like this. Hope I do not miss anything.


fanglingsu:
@rti I'm not sure about this patch. I think the `CFLAGS += ${CPPFLAGS}` is essential for package maintainers, because this allows to kick in own CPPFLAGS. Now the users CPPFLAGS are ignored completely.

rti:
Ah I see. Ok. Never mind then!

PR #340, opened 2017-04-10
Ensures correct initial input GtkTextView size

rti:
This patch is required to apply the user style defined font size on the GtkTextView input. Without it, the font size is updated on first user input, leading to a sudden unpleasant widget size and layout change if the user style font size differs from the default setting.

Unfortunately, this requires a deprecated GTK+ call. A GTK+ bug has been filed: https://bugzilla.gnome.org/show_bug.cgi?id=781158

See the diff for more details.


fanglingsu:
Thank you for the patch!

Issue #346, opened 2017-04-11
HTML from stdin (vimb -)


rti:
Allow to start vimb reading html from stdin by ```vimb -```

Implemented in vimb2 already. The code [1] can be used as a reference.

[1] https://github.com/fanglingsu/vimb/blob/master/src/main.c#L1751

PR #368, opened 2017-04-14
HTML from stdin (#346)

rti:
This patch adds reading html from stdin as already implemented on the master branch. (Issue #346)


fanglingsu:
Thank you for this patch.

Issue #370, opened 2017-04-17
incsearch does not work on clipboard paste


rti:
```incsearch``` is broken when the search string is pasted from clipboard.

### Steps to reproduce
- open a webpage
- copy a text from the current webpage to the clipboard
- activate incsearch, type ```:set incsearch=true```
- type ```/``` to start searching
- paste the text from the clipboard using ```<S-Insert>``` or GtkTextView context menu

### Expected behaviour
- matches on the webpage are highlighted

### Actual behaviour
- nothing is highlighted

Issue #369, opened 2017-04-17
mode input_changed not called


rti:
The ```Mode``` interface defines a function called ```input_changed```. The ```ex``` mode also implements it. Though, the ```Mode```'s ```input_changed``` function is never called.

PR #371, opened 2017-04-17
Fix incsearch paste

rti:
Fix for #370, on the way, #369 was fixed as well. 

Also, adds two TODOs for issue discovered. Please see the individual commits for further details.


fanglingsu:
Thanks for the patch. I kep it like it is, because removing the ifdef would introduce other issues.

PR #384, opened 2017-05-04
fixes download file name postfix (closes #383)

### LINKED ISSUE #383: Download file name postfix
When downloading the same file several times, the current implementation appends only "%u" before the file extension. This creates weird results when downloading e.g. ```program-1.4.tar.gz```. The second file downloaded would be called ```program-1.42.tar.gz```.

I suggest to insert an underscore before the counter. So the second downloaded file would be called ```program-1.4_2.tar.gz```.

rti:
see #383


fanglingsu:
Thank you, the underscore seems to be essential in some cases.

Issue #347, opened 2017-04-11
External editor


rti:
Allow editing texts (e.g. from ```<textarea>```) in external editor (e.g. vim (: ).

Implemented in vimb2 already. The code [1] can be used as reference.

[1] https://github.com/fanglingsu/vimb/blob/master/src/input.c#L105

hsoft:
I hadn't noticed this issue before and thought that it was part of #349, but as I wrote in my pR at #386, I excluded it from my work anyway...

rti:
looking into it

rti:
Draft implementation ready, ported from vimb2, using js calls from main app instead of webkit2 API for simplicity. Still requires some fixes and cleanup.

https://github.com/fanglingsu/vimb/compare/webkit2...rti:347_external_editor

fanglingsu:
@rti Looks promising. But we need a solution how to get the current form field text into the spawned editor. So I thinks we have to call the webextension to return the text from the form field and to disable the field. After editing the file we have to clal the webextension to put in the text into the right form field. This would require the webextension to keep the form field element stored so that the text can be written into the right one.

rti:
@fanglingsu this is already working. Everything you mentioned is done via JS, disabling, enabling, getting setting the value. To store the currently focussed element, it uses the same JS variable already used for input field tabbing.

The only shortcoming I can currently think of is the issue with IFrames you pointed out earlier (#328).

fanglingsu:
@rti Oh, sorry for the clutter, I didn't read the code very carefully. Yes IFrame are an issue, but I presume that we're allowed to access frames out of the same origin via the webextension.

rti:
@fanglingsu Hmm, so you think we have to take the webextension path in the end? Would you still accept a merge request implementing the current approach? Maybe it's better than nothing? I could look into the webextension thing in a next step then.

fanglingsu:
I want to have the feature and the current approach seem a good way to get it. I personally do not understand the whole dbus stuff and there seem to be many unresolved corenercases at the time. A Pull-Request is pretty welcome!

PR #386, opened 2017-05-05
Re-implement hints into webkit2 (#349)

### LINKED ISSUE #349: Hinting
Allow to highlight clickable elements and assign keyboard shortcuts to them.

Was implemented in vimb2 already. The code should be used as a reference.

There are some insights available:
- https://github.com/fanglingsu/vimb/issues/337#issuecomment-292942461
- https://github.com/fanglingsu/vimb/issues/337#issuecomment-292946599

hsoft:
This PR brings back the old `hints.c` from the `master` branch and adapts it to the webkit2 branch. This fixes #349. From my tests, most of the things work fine, with some notable exceptions:

### Open in a new window

As I wrote at https://github.com/fanglingsu/vimb/issues/349#issuecomment-298469136 , I couldn't get the old trick (dynamically adding `target="_blank"` to the element before simulating the click) to work on webkit2 because now, when we try to set `target` to `_blank` in JS, it sets it to `null`. I think that it was done to address the [window opener vulnerability](https://mathiasbynens.github.io/rel-noopener/).

I went with a simpler approach: `window.open()`. It works for simple links, but for more complex elements (buttons that trigger JS when clicked), that's not going to work. When that happens (when there's no `href` attribute), we simply revert to the `;o` behavior.

### Editor command (;e)

I haven't implemented it. I'm already knee-deep in complexity I don't master well, I'd be more comfortable implementing this after a first merge of this original code.

EDIT: oh, I see that it was *already* part of a separate issue (#347)

### Follow link under search pattern

This is unrelated to hinting, but was implemented in `hints.c` in the `master` branch. I have commented this part and suggest that we re-implement it separately, in another unit.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -130,7 +130,10 @@ gboolean command_yank(Client *c, const Arg *arg, char buf)
     g_assert(c);
     g_assert(arg);
     g_assert(c->webview);
-    g_assert(arg->i == COMMAND_YANK_URI || arg->i == COMMAND_YANK_SELECTION);
+    g_assert(
+        arg->i == COMMAND_YANK_URI ||
+        arg->i == COMMAND_YANK_SELECTION ||
+        arg->i == COMMAND_YANK_ARG);
 
     if (arg->i == COMMAND_YANK_URI) {
         if ((uri = webkit_web_view_get_uri(c->webview))) {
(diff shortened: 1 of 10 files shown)


fanglingsu:
There where some errors during hinting. I've fixed some issues, but I assume that the hinting does not work very well at the time. So I've merge your pull request into hints branch. So we can check the hinting there and merge it into webkit2 if it works properly.
To fix some issues with backspace I switched the hint js processing to syncronous dbus calls. On my system this worked, but in case we ran into performance isuues, we have to think about a solution which uses async calls by default and syncronous calls only when this is really required.
@hsoft Thank you for your hard work on the hinting.

hsoft:
I see from your commits that there's indeed many problems that still had to be fixed. I hadn't notice them!

PR #388, opened 2017-05-06
Fix build on FreeBSD

mike402:
Evening,

1. Building with _XOPEN_SOURCE=500 also needs __BSD_VISIBLE.

2. FreeBSD sed behaves differently.

I am currently updating the FreeBSD port of webkit2, then I will be able to create a port of the new vimb.

Cheers,
Mike


fanglingsu:
@mike402 Thank you for the patch. I've sqashed the two commits together.

iamleot:
Hello Michael and Daniel,

Daniel Carl writes:
> @mike402 Thank you for the patch. I've sqashed the two commits together. 

Doesn't the several `-e` (in the sed(1) incantation) needs to be
separated from the command string with a blank? (like the first
command?, i.e. `-e 'foo' vs -e'foo'`)

fanglingsu:
@iamleot I think it makes no difference in this case.

iamleot:
Daniel Carl writes:
> @iamleot I think it makes no difference in this case.

Whoops, sorry for the noise, you're right!

PR #389, opened 2017-05-09
hints: open links in async mode

hsoft:
When firing hints, open them (either by simulating a click or through
`window.open()`) in async mode with `window.setTimeout()`. If we don't
do that, the `EvalJS` dbus call will itself timeout and we'll end up
in an inconsistent state, that is, not back to Normal mode.

ref #349

### DIFF
--- a/src/scripts/hints.js
+++ b/src/scripts/hints.js
@@ -384,15 +384,23 @@ var hints = Object.freeze((function(){
 
     /* internal used methods */
     function open(e, newWin) {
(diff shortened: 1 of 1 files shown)


fanglingsu:
Thank you for the patch, nice work. I've notice the issue too, but I could not figure out why this appeared. My approach was to timeout the dbus call because without the timeout the action was never done. Your fix works perfect, even when I don't understand why the JS blocked the dbus call on `fire()`.

PR #390, opened 2017-05-09
hintElem opacity for non opaque elements

rti:
Forces the hintElem to be fully opaque, even when attached to non opaque
elements (e.g. ```{opacity:0.5}```). #349


fanglingsu:
Thank you for the patch.

PR #394, opened 2017-05-11
implement external editor

rti:
Implements the external editor for form elements.

To keep the code simple, this implementation is done via JS without touching the webextension and it's interface.

Current limitations:
 - form elements located in iframes are not supported
 - content editable elements are not supported


fanglingsu:
Thank you for your work! Pretty nice.

PR #400, opened 2017-05-18
Eliminate dead code from main.c

Ferenc-:
1. Value stored in 'num' in line 161 was never read
   since it was unconditionally overwritten immediately in 166.

2. Value stored in 'suffix' in line 145 was never read
   since it was overwritten in both branches of 'if(dot_post)'.

3. Value stored in 'req' in line 1157 was never read
   if the branch of the first if statement is taken.
   Storing has been placed int to the second if statement,
   where it is actually used.


fanglingsu:
Thank you for your investigation and contribution!

PR #401, opened 2017-05-18
fix serif-font setting name to match manpage

rti:


fanglingsu:
Well you are very attentive.

PR #404, opened 2017-05-20
Script cleanups

pks-t:
This includes one bugfix where we'd add user scripts multiple times to the user content manager and a cleanup, removing the unused hints.js.h file.


fanglingsu:
Thank you for these patches! Nice work!

PR #405, opened 2017-05-20
setting: expose hardware acceleration settings

pks-t:
Previous to the migration to the WebKit 2 API, we were exposing settings
to enable hardware acceleration for compositing via
"accelerated-compositing". With WebKit 2, there are two relevant
settings to enable compositing. The "hardware-acceleration_policy" is
used to determine when the rendering process should actually use
hardware acceleration. "accelerated-2d-canvas" enables hardware
acceleration for some 2D canvas content if WebKitGTK+ is built with
Cairo and its unstable CairoGL API.

Thes patch exposes both of these settings via vimb's settings engine.


fanglingsu:
Thank you!

PR #406, opened 2017-05-20
scroll: fix percent-based scrolling

pks-t:
Scrolling via ^D and ^U should scroll by a percentage of the client's
current viewport. We do this in JavaScript via the client height of the
document's element. Accidentally, it may hold the desired value in some
cases, but in general it only holds the total height of the element
itself. What we desire instead is the window's height, which can be
retrieved via `Window.innerHeight`. Like this, our calculations are not
based upon the webpage but instead on the browser window, fixing
scrolling on some pages.

### DIFF
--- a/src/scripts/scroll.js
+++ b/src/scripts/scroll.js
@@ -2,7 +2,7 @@ function vbscroll(mode, scrollStep, count) {
     var w = window,
         d = document,
         x = y = 0,
-        ph = d.documentElement.clientHeight,
+        ph = w.innerHeight,
         c = count||1,
         rel = true;
     switch (mode) {


pks-t:
The webpage I noted the bug on was https://webkitgtk.org/reference/webkit2gtk/stable/WebKitSettings.html, by the way. Without this patch, ^D and ^U were scrolling half the page's height, not the client's height.

fanglingsu:
Oh, there is another scroll issue on the webkitSettings page. The scroll indicator on the right of vimb's statusbar does not update.  But this has nothing to do with you patch, but maybe you can check the webextension logic in `on_document_scroll`. I assume I've done another stupid mistace, but can't see it by myself.

PR #416, opened 2017-06-12
improve hint char selection

rti:
Improves the hint char selection, especially for ```hint-keys``` settings other than the default (e.g. ```set hint-keys=asdfg;lkjh``` as suggested in the man page)

### Before
Until now the ```hint-keys``` string was used in a way that the default ```0123456789``` string generated nice decimal numbers as expected: ```1 2 3 4 5 6 7 8 9 10 11 12 13 14 15```

But setting it to something different, such as ```12345```, if you want to use only the left hand to select hints, generated somewhat strange results: ```2 3 4 5 21 22 23 24 25 31 32 33 34 35 41``` 

Same happened with the suggested home row setting from the man page: ```s d f g ; l k j h sa ss sd sf sg s;```

The first char, the ```1``` or ```a``` was never used as a leading hint character. This actually wasted it and created unnecessarily long hints.

### This patch
- Makes use of the first character as a leading hint char as well
- Therefore creates somewhat shorter hint strings
- Stays very close to the default behaviour when using the default ```hint-keys``` setting

#### The new default 1234567890
```1 2 3 4 5 6 7 8 9 0 11 12 13 14 15``` 

#### Left hand mode 12345
```1 2 3 4 5 11 12 13 14 15 21 22 23 24 25```

#### Home row mode asdfg;lkjh
```a s d f g ; l k j h aa as ad af ag```

### DIFF
--- a/src/scripts/hints.js
+++ b/src/scripts/hints.js
@@ -5,8 +5,8 @@ var hints = Object.freeze((function(){
         docs       = [],               /* hold the affected document with the start and end index of the hints */
         validHints = [],               /* holds the valid hinted elements matching the filter condition */
         activeHint,                    /* holds the active hint object */
-        filterText = "",               /* holds the typed filter text */
-        filterNum  = 0,                /* holds the numeric filter */
+        filterText = "",               /* holds the typed text filter */
+        filterKeys = "",               /* holds the typed hint-keys filter */
         /* TODO remove these classes and use the 'vimbhint' attribute for */
         /* styling the hints and labels - but this might break user */
         /* stylesheets that use the classes for styling */
@@ -30,7 +30,7 @@ var hints = Object.freeze((function(){
             this.label.style.display = "";
             this.e.classList.add(hClass);
 
-            /* create the label with the hint number */
+            /* create the label with the hint number/letters */
             var text = [];
             if (this.e instanceof HTMLInputElement) {
                 var type = this.e.type;
@@ -66,7 +66,7 @@ var hints = Object.freeze((function(){
         hints      = [];
         validHints = [];
         filterText = "";
-        filterNum  = 0;
+        filterKeys = "";
     }
 
     function create() {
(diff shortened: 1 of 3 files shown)


rti:
Just realized, this patch breakes ```set hint-number-same-length=true```.
So it should be considered work in progress atm. Sorry for the noise.

fanglingsu:
@rti Thank you! Nice investigation - this is a pretty nice patch. I'll keep it open as long as the `hint-number-same-length` does not work. So we can merge a working state.

fanglingsu:
@rti Do you still work on this, or should I try to fix the hint-number-same-length issue?

rti:
@fanglingsu Still working on it. Patch is almost ready. Just a limited amount of time on my side.  ETA before next week.

fanglingsu:
@rti Ok! I'm looking forward.

rti:
@fanglingsu If you are very curious, I am working on it over here https://github.com/fanglingsu/vimb/compare/master...rti:improve_hint_char_selection_wip

fanglingsu:
@rti Nice with manual test case. I think you don't need to link to a real page in the test you can link to the file itself `href="./"` or `href="#{num}"` in case your are interested if the hint was fired properly.

rti:
Ah yes. Good point.

rti:
Do you want the test case included in the merge request?

fanglingsu:
Yes, having the test would be nice and we can modify it later to test other aspects of the hinting too. I think it's a good decision to have some test pages we can refern too in case of bugs.

rti:
Ok, what I initially thought to be an off-by-one error turned out to be a bit more complex than expected. Anyway. Here it is now.

fanglingsu:
In case the default hint keys are used the 10th hint becomes the label '0' instead of the natural '10', do you have any idea hove we can also get nice decimal numbering `1 2 3 4 5 6 7 8 9 10 11`?

fanglingsu:
I've shortened some of the variabled because the JS is compiled into vimb so we should try to kepp it short until we can shorten minify it automatically.

PR #441, opened 2017-10-01
Rename hint-number-same-length into hint-keys-same-length #425

### LINKED ISSUE #425: Rename hint-number-same-length to hint-keys-same-length
The setting name `hint-number-same-length` is a little missleading because the hint labels are not inevitably numbers. To keep it close to the `hint-keys` setting we should rename this to `hint-keys-same-length`.

yblein:
Just a small commit to get started with vimb codebase. It renames `hint-number-same-length` into `hint-keys-same-length` to fix #425.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   hinting is done might be security relevant.
 * Allow extended hints mode also for open `g;o` to allow the user to toggle
   checkboxes and radiobuttons of forms.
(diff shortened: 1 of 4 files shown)


yblein:
Just an afterthought: maybe `hint-labels-same-length` would be more appropriate?

fanglingsu:
Thank you for the patch. I think I'll create a 'next' branch in the next days and merge this into it. Because I don't want to introduce such renaming into the current release.

fanglingsu:
Thank you for your work. I've merged this right into the master branch.

Issue #350, opened 2017-04-11
Completion of augroup commands


rti:

fanglingsu:
Depends on #356.

Issue #351, opened 2017-04-11
Completion of autocmd commands


rti:

fanglingsu:
Depends on #356.

PR #442, opened 2017-10-01
Reintroduce autocmd and augroups. Fix #356

### LINKED ISSUE #356: Autocommands and augroups
Implemented in vimb2 already. The code [1] can be used as a reference. The manpage as a spec [2].

[1] https://github.com/fanglingsu/vimb/blob/master/src/autocmd.c
[2] https://fanglingsu.github.io/vimb/man.html#Automatic_commands

yblein:
My migration to vimb3 went rather smoothly but I was missing auto commads. I mostly use them as userscripts (with the event `LoadFinished`).

This PR reintroduces autocmd and augroups almost as they were in vimb2. The commit seems big but most of the code is in fact leveraged directly from vimb2, and should work flawlessly.

The changes compared to vimb2 are in the events:
- `LoadProvisional` is renamed into `LoadStarted` for coherence
- `LoadFirstLayout` is removed as it seems webkit2 does not expose this event ([list here](https://webkitgtk.org/reference/webkit2gtk/unstable/WebKitWebView.html#WebKitLoadEvent))
- `LoadFailed` is removed for simplicty, but could be reintroduced later
- `DownloadStart` is renamed into `DownloadStarted` and is now triggered right *after* a download start.

I would also like to note that the autocmd interface is quite complex. I wonder if all the variants of `:autocmd` are really necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   bookmarks without loading the page first.
 * Refresh hints after scrolling the page or resizing the window which makes
   extended hint mode more comfortable.
+* Reintroduce the automatic commands from vimb2. An automatic command is
+  executed automatically in response to some event, such as a URI being opened.
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


fanglingsu:
Whow thank you! I thought that this would not be as easy with webkit2. I'll merge this in the next days. Pretty nice work!

yblein:
I believe this PR also closes #350 and #351.

fanglingsu:
Thank you for the patch. This works pretty nice.

PR #448, opened 2017-10-28
n/N restart the previous search if none is active

yblein:
vimb no longer ignores n/N if no search is active. Instead, it restarts
a search with the same query that was used in the last search, just like
Vim does. This is very convenient when searching for the same query in
different pages.

### DIFF
--- a/src/command.c
+++ b/src/command.c
@@ -22,6 +22,7 @@
  * together.
  */
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


fanglingsu:
@yblein Thank you for this patch. This is the behaviour of vim and the webkit2 version of vimb. I don't know why I missed this during the migration.

PR #451, opened 2017-11-07
Added stylesheet file setting

sigzegv:
I just wanted to show this update that allows to use a custom stylesheet in config file or autocommands. I'm not sure if this is the correct way, and I'm not sure about naming conventions. Tell me about any needed modifications.

With that I wonder if the stylesheet boolean parameter is still necessary, because having a definder stylesheet file or not makes  same "boolean" purpose.

### DIFF
--- a/src/setting.c
+++ b/src/setting.c
@@ -27,6 +27,7 @@
 #include "setting.h"
 #include "scripts/scripts.h"
 #include "shortcut.h"
+#include "util.h"
 
 typedef enum {
     SETTING_SET,        /* :set option=value */
@@ -60,6 +61,7 @@ static int input_autohide(Client *c, const char *name, DataType type, void *valu
 static int internal(Client *c, const char *name, DataType type, void *value, void *data);
 static int headers(Client *c, const char *name, DataType type, void *value, void *data);
 static int user_scripts(Client *c, const char *name, DataType type, void *value, void *data);
+static int style_sheet_file(Client *c, const char *name, DataType type, void *value, void *data);
 static int user_style(Client *c, const char *name, DataType type, void *value, void *data);
 static int statusbar(Client *c, const char *name, DataType type, void *value, void *data);
 static int tls_policy(Client *c, const char *name, DataType type, void *value, void *data);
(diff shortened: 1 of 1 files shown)


sigzegv:
the link hinting doesn't work with this code, don't really found why yet.. Any ideas welcomed.

src/setting.c
```diff
@@ -693,6 +696,32 @@ static int user_style(Client *c, const char *name, DataType type, void *value, v
     return CMD_SUCCESS;
 }
 
+static int style_sheet_file(Client *c, const char *name, DataType type, void *value, void *data)
+{
+    char *path;
+    gchar *source;
+
+    path = util_get_config_dir();
```
fanglingsu: This variable has to be g_freed after usage.

src/setting.c
```diff
@@ -693,6 +696,32 @@ static int user_style(Client *c, const char *name, DataType type, void *value, v
...
+    ucm = webkit_web_view_get_user_content_manager(c->webview);
+
+    WebKitUserStyleSheet *style;
+    g_file_get_contents(vb.files[FILES_USER_STYLE], &source, NULL, NULL);
+    style = webkit_user_style_sheet_new(
+        source, WEBKIT_USER_CONTENT_INJECT_ALL_FRAMES,
+        WEBKIT_USER_STYLE_LEVEL_USER, NULL, NULL
+    );
+
+    webkit_user_content_manager_add_style_sheet(ucm, style);
```
fanglingsu: I think you have to remove previouly set style sheets. Else the styles will be merged together internally.

src/setting.c
```diff
@@ -693,6 +696,32 @@ static int user_style(Client *c, const char *name, DataType type, void *value, v
...
 }
 
+static int style_sheet_file(Client *c, const char *name, DataType type, void *value, void *data)
+{
+    char *path;
+    gchar *source;
+
+    path = util_get_config_dir();
+
+    vb.files[FILES_USER_STYLE] = util_get_filepath(path, value, FALSE);
```
fanglingsu: Note that this contains already allocated memory and has to be freed before assigning new.

fanglingsu:
> the link hinting doesn't work with this code, don't really found why yet.. Any ideas welcomed.

This is because the logic to add the hinting style sheet is added in `user_style` function to make sure the style sheet is added again after changeing the user style sheet.

Issue #452, opened 2017-11-07
Hide HTTP basic auth credentials in url bar


rudism:
When HTTP basic auth credentials are supplied in a url (either typed in or saved in a bookmark), it would be nice if the credentials were not displayed in the url bar at the bottom.

For example, instead of displaying `https://username:password@domain.com/whatever` it would strip the sensitive information and just display `https://domain.com/whatever`.

fanglingsu:
@rudism Thank you for this suggestion. I've never used credential crafted URL so I haven't noticed this securitiy risk.

fanglingsu:
We forgot to close this issue.

PR #453, opened 2017-11-10
Hide username:password@ in urlbar

rudism:
My possibly naive attempt at addressing #452.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -91,6 +91,7 @@ static gboolean quit(Client *c);
 static void read_from_stdin(Client *c);
 static void register_cleanup(Client *c);
 static void update_title(Client *c);
+static gchar *sanitize_uri(const gchar *uri);
 static void update_urlbar(Client *c);
 static void set_statusbar_style(Client *c, StatusType type);
 static void set_title(Client *c, const char *title);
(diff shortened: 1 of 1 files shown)


rudism:
I switched `sanitize_uri` to use `SoupURI` as you suggested, and you are right&mdash;manually removing the credentials seems unnecessary as `soup_uri_to_string` already strips the password. The resulting displayed uri looks like `http://username@domain.com/path`.

rudism:
I think it might be worth considering the same sanitization for bookmark and history autocompletion as well as the urlbar, but I haven't familiarized myself enough with the code to know how hard that would be.

fanglingsu:
@rudism Thank you for the patch. I've moved the `sanitize_uri` to util.c so we can use this later from other files too. And as you mentioned there where a lot of other parts where the password still appeared. So I decided to sanitize the uri as soon as possilbe in the processing. So now the password is stripped in the title show if the HTML-Title is not already present, in the register '%' in the env variable $VIMB_URI (`:sh echo "$VIMB_URI"`) and newly added bookmarks.
But I've not added any code to snaitize the already present URI in the hostory and bookmark files.

PR #469, opened 2018-01-10
Fix man path

aSourceFish:


fanglingsu:
@aSourceFish I do not merge this. As I see there are two common possible locations for the man pages. One that is already used (the linux way) and the other what you want to use (bsd way). I don't want to break existing packages by changing the default location of the installed files. So I suggest you to patch this single line for yourself before install.

PR #475, opened 2018-01-31
history: add history item as soon as the load is committed

pks-t:
When changing websites, a history item for the new website is only being
added as soon as loading it has been finished. This can lead to the
situation where you start on site A, navigate to site B, click on a link
C before the load finished, but land on page A again as soon as you
navigate back.

As the described behavior is really confusing in some cases and is
actively loosing intermediate data, we should add the history item a bit
earlier. To avoid adding calls which immediately redirect to somewhere
else, a good place to add it is the LoadCommitted hook. It fires after
the final redirect as soon as initial data from the site has been
loaded.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1313,6 +1313,10 @@ static void on_webview_load_changed(WebKitWebView *webview,
              * in case a link was fired from highlighted link. */
             command_search(c, &(Arg){0, NULL}, TRUE);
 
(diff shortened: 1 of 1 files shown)


fanglingsu:
Thank you for your work. I've merge this because I thought this is the right approach to add the history items. But after testing some more I found out that the title is never available on the `LoadCommitted`. So the history would be unusable for some users that like to see the title in completion.

PR #488, opened 2018-05-28
Remember element id when returning from external editor

svensp:
Problem I'm trying to fix:
When editing an input via external editor and returning to vimb to look up some information, like a previous forum post, it sometimes got stuck on a different input element, writing the editor text to that elements value even if it did not posses one(like a select list)
This also meant the original element was not enabled again so it was not possible to return without using the web inspector.

Technical change:
Instead of relying on `vimb_input_mode_element` to still point to the correct element when the editor returns it passes the element id to the child watcher in the data.

It actually has a few more benefits than what I was trying to achieve. It is even possible to browse to different pages with the editor open as long as you return to the page with the original input element before closing the editor(and the id has not changed for some reason).
Also you can open more than one editor window at the same time for different input elements, althougth I can't imagine a case where that'd be useful.

### DIFF
--- a/src/input.c
+++ b/src/input.c
@@ -171,7 +171,7 @@ VbResult input_open_editor(Client *c)
     EditorData *data = g_slice_new0(EditorData);
     data->file = file_path;
     data->c    = c;
-	 data->element_id   = g_strdup(id);
+    data->element_id   = g_strdup(id);
     g_child_watch_add(pid, (GChildWatchFunc)resume_editor, data);
 
     return RESULT_COMPLETE;
@@ -181,6 +181,7 @@ static void resume_editor(GPid pid, int status, EditorData *data)
 {
     char *text, *escaped;
     char *jscode;
+    char *jscode_enable;
 
     g_assert(pid);
     g_assert(data);
@@ -204,12 +205,11 @@ static void resume_editor(GPid pid, int status, EditorData *data)
         }
     }
 
-	 char *jscode_enable = g_strdup_printf(
-				 "document.getElementById(\"%s\").disabled=false;"
-				 "document.getElementById(\"%s\").focus()"
-			 , data->element_id, data->element_id);
-    ext_proxy_eval_script(data->c,
-			 jscode_enable, NULL);
+    jscode_enable = g_strdup_printf(
+                "document.getElementById(\"%s\").disabled=false;"
+                "document.getElementById(\"%s\").focus()"
+                 , data->element_id, data->element_id);
+    ext_proxy_eval_script(data->c, jscode_enable, NULL);
 
     g_unlink(data->file);
     g_free(data->file);
--- a/src/input.c
+++ b/src/input.c
@@ -187,6 +187,7 @@ static void resume_editor(GPid pid, int status, EditorData *data)
     g_assert(data);
     g_assert(data->c);
     g_assert(data->file);
+    g_assert(data->element_id);
 
     if (status == 0) {
         /* get the text the editor stored */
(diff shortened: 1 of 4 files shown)


fanglingsu:
@svensp Thank you for the patch. I'll look through it soon. But this would only work in case the element has an id set. I like the idea, but I'm not sure if we can identify the element properly in the DOM if it has no id.

svensp:
That's indeed a problem.

I'd imagine that could be solved with a fallback when the id is either NULL or an empty string.
The fallback would be similar to `vimb_input_mode_element` but instead using a `Map` with the process' pid as the key.(Simply because the GPid is already passed to editor_resume and does not need another data field.)

It would retain all positives of the id variant except for being usable across site reloads / browsing then returning.

I'll play around with the thought and the concept a bit.

src/input.c
```diff
@@ -107,6 +108,8 @@ VbResult input_keypress(Client *c, int key)
 
 VbResult input_open_editor(Client *c)
 {
+    static unsigned long element_map_next_key = 1;
+    unsigned long element_map_key = 0;
```
fanglingsu: I do not see the benefit in having two variables to increment a counter. Wouln't it be enought to use only `static usigned long` and increment it?

src/input.c
```diff
@@ -136,6 +139,16 @@ VbResult input_open_editor(Client *c)
     idreturn = ext_proxy_eval_script_sync(c, "vimb_input_mode_element.id");
     g_variant_get(idreturn, "(bs)", &idSuccess, &id);
 
+    if( !idSuccess || !id || strlen(id) == 0) {
+        element_map_key = element_map_next_key++;
+        char *js_command = g_strdup_printf(
```
fanglingsu: You could also use JavaScript in `src/scripts/*.js` so this will be minified and escaped so you only need to use the constant instead of the long string. The constant is wirtten to `src/scripts/scripts.h` file.

src/input.c
svensp: The idea behind the second variable is for a clean code flow.

The default value of `0` is set during initialization.
The only logic concerning its change is within the code block that handles the special case `element does not have an id`

althought to be true to this I should have applied the same logic to the `c->element_id` assignment

src/input.c
svensp: Cool, I'll change that. I only looked at the code inside input.c and did not see you can do that.

svensp: Should I use this for all javascript within the code or just for the multi line scripts?

src/input.c
fanglingsu: It's your choice. I didn't spent much tinking about, but I think we could use a JavaScript function or two to do the dirty stuff for us. So that we only need to call these functions with the `element_map_key` as param.
It' not necessary to put more and more changes into this pull request. If it works, I'll merge it. We can than cleanup some things afterwards.

src/input.c
fanglingsu: I think it would be enough to do this for the multiliners or when the code is hard to read because of the heavy quote escaping.
If I'll have some spare time in future I'd like to move all the JavaScript snippets into one or more JavaScript Objects that are injected into the DOM so that we only need to call our little helpers instead of running scripts snippets via dbus on the webextension every time.

src/input.c
svensp: I agree. A vimb javascript object would be nice to keep the javascript organized and injected at a generalized place. On the other hand those javascript functions are also open to be called by the pages' javascript so that might be a weird security concern. Probably too small of a user base to be of any interest thought.
Anyway for now I have pushed the latest version of my code. I've been actively using it for the last few days along with my tabbed patch and I have not had any problems so far.

fanglingsu:
@svensp Thank you for your work.

PR #489, opened 2018-06-11
Clarify :set usage in man page

GKFX:
It takes a while to work out from the man page how to use `:set`. Hopefully this clears up some things.


fanglingsu:
Thank you for the patch!

PR #490, opened 2018-06-12
Allow user to hide and show the titlebar on demand. Fixes #439.

### LINKED ISSUE #439: Possible to disable window decorator?
Is it possible to turn off the decorators that comes with gtk? i find the titlebar annoying when using a tilling wm?

Thanks

GKFX:
This works for me, and since my system is pretty mainstream it should work for lots of others too. As noted in the code, it doesn't provide a way for the user to hide the titlebar if they're using a system where you have to call `gtk_window_set_decorated` before showing the window, since the config file is read too late. (Obviously the call to `ex_run_file` in `main.c` could just be moved up a few lines but since that might have broken something else I haven't done it.)
I don't know how common it is for a window manager to make this not work. Gtk has been around for a while and this might well be a caveat that's only applicable to obscure systems from 10 years ago.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1215,6 +1215,9 @@ Number of pixel vimb scrolls if 'j' or 'k' is used.
 .B serif-font (string)
 The font family used as the default for content using serif font.
 .TP
+.B show-titlebar (bool)
+Determines whether the titlebar is shown (on systems that provide window decoration). Defaults to true.
+.TP
 .B site-specific-quirks (bool)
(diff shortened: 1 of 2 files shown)


cdlscpmv:
I think you forgot about documentation.

GKFX:
Looks like I did! You should be able to see some now.

fanglingsu:
@GKFX Thank you for the patch. I could test this on the laptop of my wife which uses xfce. The decoration can there be toggled on and off.

PR #491, opened 2018-06-12
Detect localhost and IPv6 as valid URLs. Fixes #478.

### LINKED ISSUE #478: Does not recognize localhost as address.
### Steps to reproduce
:open localhost

### Expected behaviour
connect to my localhost web server

### Actual behaviour
searches for "localhost" using default search engine

GKFX:
The following commands now work:
```
:open [::1]
:open [::1]:631
:open localhost
:open localhost:631
```
This still (correctly) goes to a shortcut:
```
:open [::1:631
:open ::1]:631
```
As was already the case, some malformed URLs don't get sent to a shortcut, and end up at `about:blank`.
```
:open 127.0.0.1:foobar
:open ::1[]:631
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -353,6 +353,21 @@ void vb_input_update_style(Client *c)
 }
 
 /**
+ * Tests if a path is likely intended to be an URI (given that it's not a file
+ * path or containing "://").
+ */
+gboolean vb_plausible_uri(const char *path) {
+  if (strchr(path, ' ')) return FALSE;
+  if (strchr(path, '.')) return TRUE;
+  const char *i, *j;
+  if ((i = strstr(path, "localhost")) &&
+      (i == path || i[-1] == '/' || i[-1] == '@') &&
+      (i[9] == 0 || i[9]  == '/' || i[9] == ':')) return TRUE;
+  return ((i = strchr(path, '[')) && (j = strchr(i, ':')) && strchr(j, ']'))
+    ? TRUE : FALSE;
+}
+
+/**
  * Load the a uri given in Arg. This function handles also shortcuts and local
  * file paths.
  *
@@ -382,11 +397,9 @@ gboolean vb_load_uri(Client *c, const Arg *arg)
         rp  = realpath(path, NULL);
         uri = g_strconcat("file://", rp, NULL);
         free(rp);
-    } else if (strchr(path, ' ') || !(strchr(path, '.') ||
-          (strchr(path, '[') && strchr(path, ':') && strchr(path, ']')) ||
-          strstr(path, "localhost"))) {
+    } else if (!vb_plausible_uri(path)) {
         /* use a shortcut if path contains spaces or doesn't contain typical
-         * characters ('.', [:] for IPv6 addresses, 'localhost') */
+         * tokens ('.', [:] for IPv6 addresses, 'localhost') */
         uri = shortcut_get_uri(c->config.shortcuts, path);
     }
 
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -382,8 +382,11 @@ gboolean vb_load_uri(Client *c, const Arg *arg)
         rp  = realpath(path, NULL);
         uri = g_strconcat("file://", rp, NULL);
         free(rp);
-    } else if (strchr(path, ' ') || !strchr(path, '.')) {
-        /* use a shortcut if path contains spaces or no dot */
+    } else if (strchr(path, ' ') || !(strchr(path, '.') ||
+          (strchr(path, '[') && strchr(path, ':') && strchr(path, ']')) ||
+          strstr(path, "localhost"))) {
```
cdlscpmv: I think you also need to check that the path actually starts with the string "localhost" and not, for example, "mylocalhost".

src/main.c
```diff
@@ -382,8 +382,11 @@ gboolean vb_load_uri(Client *c, const Arg *arg)
         rp  = realpath(path, NULL);
         uri = g_strconcat("file://", rp, NULL);
         free(rp);
-    } else if (strchr(path, ' ') || !strchr(path, '.')) {
-        /* use a shortcut if path contains spaces or no dot */
+    } else if (strchr(path, ' ') || !(strchr(path, '.') ||
+          (strchr(path, '[') && strchr(path, ':') && strchr(path, ']')) ||
```
cdlscpmv: You are not checking for the relative order of `[`,`:` and `]`. Also you need to check that the address actually starts with `[`.

cdlscpmv:
Maybe it's worth looking at [libsoup API](https://developer.gnome.org/libsoup/stable/SoupURI.html) for these things. For example, `SOUP_URI_VALID()` or `SOUP_URI_VALID_FOR_HTTP()` may turn out to be useful. It would probably simplify this whole `if` construct, and the next one as well.

GKFX:
@cdlscpmv Using libsoup to validate URIs is a nice idea but it doesn't actually work very well; possibly because of a mismatch between what a software library thinks is a good URI, and what a user wants to type into a browser. These are considered URIs:
```
localhost:631
example.com:80
```
These aren't:
```
localhost
google.com
[::1]:631
```
That's using `soup_uri_new()` and testing if the result is null, not even using the stricter `SOUP_URI_VALID()` or `SOUP_URI_VALID_FOR_HTTP()` macros as you suggest. Preceding with `http://` causes libsoup to accept any arbitrary nonsense.

Ultimately the question isn't “is this input a valid URI”, it's  “is this supposed to be an URI” which is really very different. The code is always going to be using a slightly sloppy heuristic, although it's possible that what I've put is a bit too sloppy. I'll make it a bit more rigorous and put it in a function.

cdlscpmv:
Thank you for testing things out! I've never checked it myself before.

fanglingsu:
@GKFX Thank you for your work. Write if you are done with this more rigorous solution.

GKFX:
This now checks that:
* For localhost urls, localhost is preceded by `/` or `@`, or is the start of the string; and is followed by `/` or `:`, or is the end of the string.
* For IPv6 addresses, the first `[` must be followed by a `:` somewhere, followed by a `]`.
* Previous checks for spaces and dots are unchanged

This allows any sensible urls and addresses @cdlscpmv's concerns.

fanglingsu:
@GKFX Thank you! I've rename the function because the `vb_` prefix is only used for function meant to be called from outside (none static).

PR #493, opened 2018-06-16
Fix missing string.h include

infmagic2047:
It is causing an incompatible implicit declaration warning.


fanglingsu:
@infmagic2047 Thank you for the patch.

PR #496, opened 2018-06-18
Makefile: add dependency on scripts.h to input.o

pks-t:
While "input.c" includes "scripts/scripts.h", which is generated
dynamically by the build system, the Makefile does not state a
dependency of "input.c" on "scripts/scripts.h". Add the dependency to
fix broken builds.


fanglingsu:
@pks-t Thank you for the patch.

PR #498, opened 2018-06-18
Add include paths for Debian packages so Vimb builds out-of-box

GKFX:
Currently to build on a Debian system it's necessary to determine and add these lines to `config.mk` by hand; it should now be possible to just `git pull` and build.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -24,6 +24,13 @@ COMMIT := $(shell git describe --tags --always 2> /dev/null || echo "unknown")
 
 # setup general used CFLAGS
 CFLAGS   += -std=c99 -pipe -Wall -fPIC
+CFLAGS   += -I/usr/include/atk-1.0
+CFLAGS   += -I/usr/include/cairo
+CFLAGS   += -I/usr/include/glib-2.0
+CFLAGS   += -I/usr/include/gtk-3.0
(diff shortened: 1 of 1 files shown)


config.mk
```diff
@@ -24,6 +24,13 @@ COMMIT := $(shell git describe --tags --always 2> /dev/null || echo "unknown")
 
 # setup general used CFLAGS
 CFLAGS   += -std=c99 -pipe -Wall -fPIC
+CFLAGS   += -I/usr/include/atk-1.0
```
fanglingsu: Is this really needed? Note you can give the CFLAGS direct to the calls of make.
```
CFLAGS="-I/usr/include/cairo -I/usr/..." make ...
```
I think the include paths should be set by the line `CFLAGS += $(shell pkg-config --cflags $(LIBS))` at the end of the _config.mk_. You could check this by calling
```
pkg-config --cflags gtk+-3.0 'webkit2gtk-4.0 >= 2.3.5'
```

fanglingsu:
@GKFX Can you provide the output of `make options`. The required FLAGS should already be set.

fanglingsu:
Vimb compiles on ubuntu without changes, so I assume it should also under debian.

PR #499, opened 2018-06-19
Correct the man page

cdlscpmv:
Related to the introduction of the new `vimbhint` attribute.


fanglingsu:
@cdlscpmv Thank you for the patch. I didn't remenber that something like this was mentioned in the man page too.

PR #503, opened 2018-07-18
Improve scripts change detection in Makefile

hsoft:
When changing a script in `src/scripts`, `make` would need to be invoked
twice for a proper build: once for `scripts.h` and once for the `.o`
files depending on it.

This commit fixes this by always calling `scripts.subdir-all`.


fanglingsu:
@hsoft Nice investigation. But ordering the requirements for the target 'all' does not properly work, as you can see on the travis build. If Make is called to run in parallel. So I think we have to write the dependencies explicitly.

hsoft:
Hum, yes, I see. Will try to fix later today.

fanglingsu:
Maybe it was a bad decision of me to implement the Makefile in each directory separate not concerning possible dependencies between compilation target built in one directory but required in another one.

hsoft:
After some messing around, I could indeed not manage to fix the issue without removing the recursive `make` call for `src/scripts`. This commit fixes the problem, but I don't know how you feel about removing `src/scripts/Makefile`. Sure, it introduces an inconsistency in the makefiles, but it also fixes a real problem.

fanglingsu:
@hsoft Thank you for the patch. Now it's better - the `scripts` directory is only to keep the `src` directory neat. But the sources in `scripts` are only to generate dependencies for thins done in `src`. I think we can write the `script.h` file into `src` later.

PR #504, opened 2018-07-27
WIP: Convert JS to C

hsoft:
This is a proposal accompanied by a small POC: How about getting rid of the JS code and converting all to C?

I feel like the current Main -> dbus -> JS system make our code more complicated than it should. `hint.js` is a bit too complicated. I'm not sure yet, but I have the feeling that re-implementing all JS code in C would result in a simpler and more elegant code overall.

Of course, right now the dbus interfacing code is a bit messy, but I'm sure we could re-organize it into something a bit more elegant.

@fanglingsu, what do you think?

### DIFF
--- a/src/ext-proxy.c
+++ b/src/ext-proxy.c
@@ -224,6 +224,16 @@ void ext_proxy_set_header(Client *c, const char *headers)
     dbus_call(c, "SetHeaderSetting", g_variant_new("(s)", headers), NULL);
 }
 
+void ext_proxy_lock_input(Client *c, const char *element_id)
+{
+    dbus_call(c, "LockInput", g_variant_new("(ts)", c->page_id, element_id), NULL);
+}
+
+void ext_proxy_unlock_input(Client *c, const char *element_id)
+{
+    dbus_call(c, "UnlockInput", g_variant_new("(ts)", c->page_id, element_id), NULL);
+}
+
 /**
  * Call a dbus method.
  */
--- a/src/webextension/ext-dom.c
+++ b/src/webextension/ext-dom.c
@@ -33,11 +33,11 @@ gboolean ext_dom_is_editable(WebKitDOMElement *element)
 {
     char *type;
     gboolean result = FALSE;
-    
+
     if (!element) {
         return FALSE;
     }
-    
+
     /* element is editable if it's a text area or input with no type, text or
      * password */
     if (webkit_dom_html_element_get_is_content_editable(WEBKIT_DOM_HTML_ELEMENT(element))
@@ -170,6 +170,27 @@ char *ext_dom_editable_get_value(WebKitDOMElement *element)
     return value;
 }
 
+void ext_dom_lock_input(WebKitDOMDocument *parent, char *element_id)
+{
+    WebKitDOMElement *elem;
+
+    elem = webkit_dom_document_get_element_by_id(parent, element_id);
+    if (elem != NULL) {
+        webkit_dom_element_set_attribute(elem, "disabled", "true", NULL);
+    }
+}
+
+void ext_dom_unlock_input(WebKitDOMDocument *parent, char *element_id)
+{
+    WebKitDOMElement *elem;
+
+    elem = webkit_dom_document_get_element_by_id(parent, element_id);
+    if (elem != NULL) {
+        webkit_dom_element_remove_attribute(elem, "disabled");
+        webkit_dom_element_focus(elem);
+    }
+}
+
 /**
  * Indicates if the give nelement is visible.
  */
@@ -177,3 +198,4 @@ static gboolean is_element_visible(WebKitDOMHTMLElement *element)
 {
     return TRUE;
 }
+
--- a/src/webextension/ext-main.c
+++ b/src/webextension/ext-main.c
@@ -84,6 +84,14 @@ static const char introspection_xml[] =
     "  <method name='SetHeaderSetting'>"
     "   <arg type='s' name='headers' direction='in'/>"
     "  </method>"
+    "  <method name='LockInput'>"
+    "   <arg type='t' name='page_id' direction='in'/>"
+    "   <arg type='s' name='elemend_id' direction='in'/>"
+    "  </method>"
+    "  <method name='UnlockInput'>"
+    "   <arg type='t' name='page_id' direction='in'/>"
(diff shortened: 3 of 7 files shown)


src/webextension/ext-dom.c
```diff
@@ -170,10 +170,23 @@ char *ext_dom_editable_get_value(WebKitDOMElement *element)
     return value;
 }
 
+void ext_dom_focus_element_by_id(WebKitDOMDocument *parent, char *element_id)
```
fanglingsu: We should use a better name, because this does not only focus the element.

src/ext-proxy.c
```diff
@@ -224,6 +224,11 @@ void ext_proxy_set_header(Client *c, const char *headers)
     dbus_call(c, "SetHeaderSetting", g_variant_new("(s)", headers), NULL);
 }
 
+void ext_proxy_focus_element_by_id(Client *c, const char *element_id)
+{
+    dbus_call(c, "FocusElementById", g_variant_new("(s)", element_id), NULL);
```
fanglingsu: We have to check if the call to the webextension was meant to the right pageId. And there is also a check in the webextension for that. So you need to give the pageId for the call `dbus_call(c, "FocusElementById", g_variant_new("(ts)", c->page_id, element_id), NULL);`

src/webextension/ext-main.c
```diff
@@ -84,6 +84,9 @@ static const char introspection_xml[] =
     "  <method name='SetHeaderSetting'>"
     "   <arg type='s' name='headers' direction='in'/>"
     "  </method>"
+    "  <method name='FocusElementById'>"
+    "   <arg type='s' name='elemend_id' direction='in'/>"
```
fanglingsu: Parameter for the pageId is required too `<arg type='t' name='page_id' direction='in'/>`.

src/webextension/ext-main.c
```diff
@@ -419,6 +422,14 @@ static void dbus_handle_method_call(GDBusConnection *conn, const char *sender,
         }
         ext.headers = soup_header_parse_param_list(value);
         g_dbus_method_invocation_return_value(invocation, NULL);
+    } else if (!g_strcmp0(method, "FocusElementById")) {
+        g_variant_get(parameters, "(s)", &value);
```
fanglingsu: Pageid must be read from the parameters.

src/webextension/ext-dom.c
```diff
@@ -170,10 +170,23 @@ char *ext_dom_editable_get_value(WebKitDOMElement *element)
...
+{
+    WebKitDOMElement *elem;
+    GError *error = NULL;
+
+    elem = webkit_dom_document_get_element_by_id(parent, element_id);
+    if (elem != NULL) {
+        webkit_dom_element_set_attribute(elem, "disabled", "false", &error);
```
fanglingsu: In case of an error the variable `error` have to be freed by `g_error_free()`, but I assume you can give NULL instead of unused var.
But this function does not work for me. I think it would be better to use `webkit_dom_element_remove_attribute` instead, which works for me.

fanglingsu:
@hsoft You are tight the Main -> dbus -> JS is a little bit messy. And the whole dbus code does not satisfy me. I'm new to dbus and do not know how to use it right. So feel free to improve it.

I'm not sure if reimplementation of the JS code in c would look better or run better. Especially for the hinting. Vimb used in it's early days hinting implementented via webkit dom api [see changes](https://github.com/fanglingsu/vimb/compare/319160f5b4800165a8480cc2dab95d287acbc81e...df6ddbad9e73a8dc53b67de52a7d899e81814f21). OK this was webkit1, but this was damn slow if the page contained more than 50 links. As I remember the culprit was the `dom_elemen_get_bounding_rect` my attempt of something like the JS `getBoundingClientRect();`. There was no api function to get these information so we had to traverse the dom tree up to sum up the element offsets. In JS this is a really fast call.
I'm no JS programmer, but also with my little knowledge it is easy to read and to adapt.

If you have too many time, write a litte prototype for the hinting via c so we can check if the required function are now available in the dom api. But improving the communication between main and the webextension feels to be more important to me.

hsoft:
@fanglingsu you mean something like https://webkitgtk.org/reference/webkitdomgtk/stable/WebKitDOMElement.html#webkit-dom-element-get-bounding-client-rect didn't exist in webkit1 (well in fact, it seems it was added in v2.18)?

fanglingsu:
@hsoft If this exists now, we could try to migrate the logic to c. But I fear that would be a hard work.

hsoft:
I corrected my first commit.

As for the rest, I think I'll give it a try, we'll see how it looks.

hsoft:
I started working on something but I ended up stumbling on https://webkitgtk.org/2018/04/18/webkitgtk2.21.1-released.html which deprecates most of the interesting functions I was wanting to use.

It seems to me that if we're going to significantly refactor this part of the code, we might as well upon webkit-gtk's new Glib JavascriptCore API.

fanglingsu:
@hsoft Thank you for the patch.
Puh, JavascriptCore API sounds complicated to me. And I know there are some debian users that do not have bleading edge libraries so I fixed the webkit version to minimum of 2.8.x. Not shure if this is the common used version, but I don't want to to lock out users which use older webkit versions.

hsoft:
Hum, I wasn't really expecting you to merge this WIP. As a partial effort, it wasn't much of an improvement over the previous code...

fanglingsu:
@hsoft If you want to do some more usefule changes, you could implement the prev-next feature of vimb1. This was implemented in [hints.js](https://github.com/fanglingsu/vimb/blob/2.12/src/hints.js#L461-L507) and [normal.c](https://github.com/fanglingsu/vimb/blob/2.12/src/normal.c#L648-L659).

I want this nice feature back and I think this could be done in the webextension by using the DOM API.

fanglingsu:
@hsoft Are you still interested into optimization to replace JavaScript code by using the DOM API, or by cleaning up the big hints.js file?

hsoft:
@fanglingsu sorry, I got carried away. I stopped using vimb. It's the second time it happens. At first, I say "I'm going to set myself up efficiently" and then after a few months, I tell myself it's not worth it.

Sorry for the half-finished work, but to my defense, I wasn't expecting you to merge it at that point :) Feel free to revert your merge if you feel it harms the project more than helping it.

PR #517, opened 2018-09-27
Clarify proxy configuration

allanwind:
The current docs reference a non-existing proxy configuration variable, and you get an error if set.  While either HTTP_PROXY or http_proxy is accetpable environment variables the former is generally used.


allanwind:
Thank you, sir.

fanglingsu:
@allanwind Thank you for your contribution! I squashed the three commits as one onto master.

PR #519, opened 2018-10-11
Fix a copypasto regarding user-scripts option

iamleot:
The description of user-scripts option accidentally referred to
userstyles, probably due a copypasto of the option description in
the man page (or, in other words, enabling/disabling user-scripts
option really enable/disable running user scripts via user.js
file!).


fanglingsu:
Thank you!

Issue #355, opened 2017-04-11
Page marks


rti:
Maybe we change make them global (shared between instances and to work also if the page was reloaded or changed like the marks in vim)

aksr:
This is one thing I miss the most from vimb2.
Any ETA of this feature?

fanglingsu:
There are some challenges to solve to get this back into vimb. And i did not like the limitations of this feature to hold the marks only for the current opened page. If you open an other page and come back the marks where not set anymore. So I did not spend any time to get this back. If someone else would implement this I'd be pretty happy to merge this.
I've thought about saving the url and some position value in a separate file (to have some global marks) or to store the url and the position in the current instance for local marks. But the main challenge is to get the position (via webextension) and to restore them if the current url is the same as the one the mark was set for (I think location hashes '#' can be ignored for this).

fanglingsu:
Page marks are now implemented like in vimb 2 #521.

PR #521, opened 2018-10-15
Add support for marks, based on vimb-2.12.

iamleot:
These two commits, one for vimb source code, the other to update
vimb(1) man page restore the support for page marks as present in
vimb-2.12 (and also most of the code is based on it, in particular
normal_map() implementation!).

This should also partly address issue #355.


fanglingsu:
@iamleot Thank you for this patch!

PR #524, opened 2018-10-15
Document a subtlety of util_parse_expansion() and adjust test_expand_tilde_user() accordingly

iamleot:
Single ~ uses g_get_home_dir() that honor HOME environment variable.
However, ~user always consult the passwd file.

Adjust test_expand_tilde_user() to reflect this behaviour and always use
getpwnam() there (instead of g_get_home_dir() that can lead to incorrect
results if HOME is set).

This was noticed when running the `test' target on pkgsrc, where
an HOME directory is set to a fake directory.


fanglingsu:
Wow! Another nice PR of you!

PR #530, opened 2018-11-11
Fix out-of-bounds buffer access in parse_command

### LINKED ISSUE #529: Config parser segfault
Hi, thanks for writting this software. Just wanted to report a segfault I encountered while experiminting with it.

```
$ vimb --bug-info
Commit:          8f9e1b6
WebKit compile:  2.22.3
WebKit run:      2.22.3
GTK compile:     3.24.1
GTK run:         3.24.1
libsoup compile: 2.64.1
libsoup run:     2.64.1
Extension dir:   /usr/lib/vimb
```

Running on Alpine Linux Edge.

### Steps to reproduce

```
$ echo "spell-checking-languages=en,de" > conf
$ vimb -c conf
Segmentation fault (core dumped)
```

Backtrace:

```
#0  0x00007fac4acb9650 in a_crash () at ./arch/x86_64/atomic_arch.h:108
#1  0x00007fac4acb9650 in __stack_chk_fail () at src/env/__stack_chk_fail.c:17
#2  0x0000557c1967ea45 in parse_command_name (c=c@entry=0x557c1b1bc160, input=input@entry=0x7ffcf09a48e0, arg=arg@entry=0x557c1b54ad80) at ex.c:674
#3  0x0000557c1967fc93 in parse (nohist=<synthetic pointer>, arg=0x557c1b54ad80, input=0x7ffcf09a48e0, c=0x557c1b1bc160) at ex.c:583
#4  0x0000557c1967fc93 in ex_run_string (c=c@entry=0x557c1b1bc160, input=input@entry=0x557c1b21eb60 "spell-checking-languages=en,de", enable_history=enable_history@entry=0) at ex.c:506
#5  0x0000557c196802fd in ex_run_file (c=0x557c1b1bc160, filename=0x557c1b1ae160 "/tmp/conf") at ex.c:485
#6  0x0000557c19675dca in main (argc=<optimized out>, argv=<optimized out>) at main.c:1977
```

### Expected behaviour

The config file is invalid. The correct way to set spell checking languages is:

```
set spell-checking-languages=en,de
```

Nonetheless, vimb should output an error message instead of segfaulting.

### Actual behaviour

vimb segfaults.

nmeum:
Since the cmd buffer needs to be null terminated we need to reserve
space for the null byte in the buffer.

Without this change an out-of-bounds buffer access is performed if the
first word is longer than 19 chars.

Fixes #529

### DIFF
--- a/src/ex.c
+++ b/src/ex.c
@@ -657,7 +657,7 @@ static gboolean parse_command_name(Client *c, const char **input, ExArg *arg)
         /* read until next whitespace or end of input to get command name for
          * error message - vim uses the whole rest of the input string - but
          * the first word seems to bee enough for the error message */
-        for (; len < LENGTH(cmd) && *input && !VB_IS_SPACE(**input); (*input)++) {
+        for (; len < (LENGTH(cmd) - 1) && *input && !VB_IS_SPACE(**input); (*input)++) {
             cmd[len++] = **input;
         }
         cmd[len] = '\0';


fanglingsu:
@nmeum Thank you for your investigation! Great work!

Issue #76, opened 2014-04-07
Build Error in FreeBSD


xgeek-cub:
Hello;
While building in freebsd-10-stable; I get these errors both with `GTK=3` and without.

```
make: "/usr/home/avalon/build/vimb/config.mk" line 16: Missing dependency operator
make: "/usr/home/avalon/build/vimb/config.mk" line 17: Missing dependency operator
make: "/usr/home/avalon/build/vimb/config.mk" line 20: Need an operator
make: "/usr/home/avalon/build/vimb/config.mk" line 22: Need an operator
make: "/usr/home/avalon/build/vimb/config.mk" line 23: Need an operator
make: "/usr/home/avalon/build/vimb/config.mk" line 24: Need an operator
make: "/usr/home/avalon/build/vimb/config.mk" line 26: Need an operator
make: "/usr/home/avalon/build/vimb/config.mk" line 34: Missing dependency operator
make: "/usr/home/avalon/build/vimb/config.mk" line 36: Need an operator
make: Fatal errors encountered -- cannot continue
make: stopped in /usr/home/avalon/build/vimb
```

PR #538, opened 2019-01-26
Define feature macros only on Linux

mazocomp:


mazocomp:
https://github.com/cylgom/termbox-next/pull/1

iamleot:
Hello Leonid,

Leonid Bobrov writes:
> [...]
>   * Define feature macros only on Linux
> [...]

Why? (at least -D_XOPEN_SOURCE but probably also -D__BSD_VISIBLE
is honored by other operating systems as well)

mazocomp:
@iamleot Hi.

Take a look at link above: it refuses to compile at *BSD (I am running DragonFly BSD).

Also any macro starting with __ is not supposed to be used by Makefile, so having __BSD_VISIBLE as a flag is a bug (try _BSD_SOURCE instead), but I care about compilation at *BSD and Linux, I can't test at other OSes.

iamleot:
Leonid Bobrov writes:
> [...]
> Take a look at link above: it refuses to compile at *BSD (I am running DragonFly BSD).
>
> Also any macro starting with __ is not supposed to be used by Makefile, so having __BSD_VISIBLE as a flag is a bug (try _BSD_SOURCE instead), but I care about compilation at *BSD and Linux, I can't test at other OSes.
> [...]

AFAIK only FreeBSD and DragonFly BSD has that behaviour.
It is probably safer to conditionally remove the CPPFLAGS only for
them instead of enabling them only for Linux.

mazocomp:
@imleot Does the current diff look ok? I don't know how GNU make works, so I am sorry for possible mistake.

iamleot:
Leonid Bobrov writes:
> @imleot Does the current diff look ok? I don't know how GNU make works, so I am sorry for possible mistake.

I think that the ifneq should be replaced with something like that:

    ifeq "$(findstring $(OS),FreeBSD DragonFly)" ""
    ...
    endif


(Also, if possible it is usually better to add a rationale in the
commit message (why a change is needed), but beware!, I'm not a
vimb developer! :))


Thank you!

fanglingsu:
@mazocomp Both defines where added to get rid of FreeBSD compile errors #76. So I think your patch should work also on gnu systems.

mazocomp:
@fanglingsu my patch works for both FreeBSD and DragonFly BSD (note that Vimb failed to compile at DragonFly BSD without this patch), I didn't test it at OpenBSD and NetBSD yet, but it should work there too. And remember: It's a bug to define macros in compiler which start with __, also please see the first link I sent in this PR: 
https://github.com/cylgom/termbox-next/pull/1

PR #542, opened 2019-03-05
Use man page date instead of build date

bmwiedemann:
Use man page date instead of build date
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Also use UTC to be independent of timezone.

note: This date call works with GNU date and BSD date (e.g. macOSX).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ install: all
 	install -d $(MANPREFIX)/man1
 	@sed -e "s!VERSION!$(version)!g" \
 		-e "s!PREFIX!$(PREFIX)!g" \
-		-e "s!DATE!`date +'%m %Y'`!g" $(DOCDIR)/vimb.1 > $(MANPREFIX)/man1/vimb.1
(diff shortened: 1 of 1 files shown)


iamleot:
Hello Bernhard, 

Bernhard M. Wiedemann writes:
> Use man page date instead of build date
> in order to make builds reproducible.
> See https://reproducible-builds.org/ for why this is good.
>
> Also use UTC to be independent of timezone.

While I think that's good to have, please note that `date -r' semantic
is not portable (I think it is supported only by GNU date and maybe
FreeBSD date(1)).

I don't know if it will add possible more maintainance overhead
but maybe just bumping the date manually like vimb version for
every stable release is another way to address possible non reproducible
builds.

bmwiedemann:
GNU date and BSD date cover a lot of platforms. I know Solaris had a date without -r - is that a target for vimb? I now added a fallback there, so even on Solaris it will not be worse than before.

iamleot:
Bernhard M. Wiedemann writes:
> GNU date and BSD date cover a lot of platforms. I know Solaris had a date without -r - is that a target for vimb? I now added a fallback there, so even on Solaris it will not be worse than before.

At least NetBSD and OpenBSD date(1) have a different semantic for
`-r'.  And, IMHO it would be better to address that unconditionally
for every platform instead of having a different date depending on
the platform.

fanglingsu:
@bmwiedemann Thank you for providing the patch. I've never thought about reproducible builds before.

Issue #546, opened 2019-03-14
inoremap prevents character inputing after timeout


noboruma:
vimb, version 3.3.0-21-ga884a81

### Steps to reproduce
I have the following in my config:
inoremap jj \<Esc\>j

### Expected behaviour
If I input j and wait a bit, I would expect j to appear in the input.

### Actual behaviour
j never shows up. jj works, j does nothing.

fanglingsu:
@noboruma Thank you for this report. We had a similar issue year before #258 which was solved by a pull request. But I can't find any code of this now. This might have been dropped during migration to webkit2.

noboruma:
@fanglingsu Thanks for the reply! Good to know this issue was already raised, I will see if I can get it working and make a new PR.

noboruma:
#547 is waiting for approval, I am closing this issue.

PR #547, opened 2019-03-14
Fix for #546

noboruma:
Fix for #546 based on Michael Mackus' work.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -2,7 +2,7 @@ ifneq ($(V),1)
 Q := @
 endif
 
-PREFIX           ?= /usr/local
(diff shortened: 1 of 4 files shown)


noboruma:
Tested on my end, it works fine :-)
@fanglingsu thanks again for pointing me at the fix, and for vimb!

fanglingsu:
@noboruma Thank you for providing this patch. I've applied some changes to fix warning shown on cli and to use the glib memory allocation function. And moved the whole stuff into map.c because we need it only there and can save some lines of code.

PR #561, opened 2019-05-20
Fix NULL ret of util_sanitize_uri

tharvik:
* in `util_sanitize_uri`,  `webkit_uri_for_display` can return NULL, so failing early
* `util_sanitize_uri` can return NULL, so  adding check of return value before using it

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1445,7 +1445,9 @@ static void on_webview_load_changed(WebKitWebView *webview,
             /* update load progress in statusbar */
             c->state.progress = 0;
             vb_statusbar_update(c);
-            set_title(c, uri);
+            if (uri) {
+                set_title(c, uri);
+            }
             /* Make sure hinting is cleared before the new page is loaded.
              * Without that vimb would still be in hinting mode after hinting
              * was started and some links was clicked my mouse. Even if there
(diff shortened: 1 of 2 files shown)


fanglingsu:
@tharvik Thank you for this patch. I've seen the "... (null)" in title someday in the past but forgot to investigate into this.

tharvik:
Oups, just reread the commit, there is a double free in `on_webview_notify_uri`, when calling a second time after en empty URL.

fanglingsu:
@tharvik I'm not sure what to do in this case. We have two options.
- Free the `c->state.uri` only if we get a new sanitized uri. But this would keep a possible wrong uri set as env `VIMB_URI`.
- Set `c->state.uri` to empty string. This might also be uncommon.

tharvik:
The first one seems better, as I think that an empty URL is not really relevant for `VIMB_URI`, and previsiously, it would have been `(null)` if I'm correct.

btw, there might be something deeper than this quick fix, such as why we got a empty URL; maybe there is a better place to fix it?

fanglingsu:
I'll change the `util_sanitize_uri` to not return NULL in case the sanitization in `webkit_uri_for_display` fails and follow epiphany and return the given url in these rare cases https://github.com/GNOME/epiphany/blob/6141d65eee4e74ca8904ed9cdd77dc0c7c892f18/lib/ephy-uri-helpers.c#L260-L270.

fanglingsu:
@tharvik I've change the `util_sanitize_uri` to return given string in case webkit turned none empty string into NULL in case an error occurred.

tharvik:
Seems to be working here, thanks :)

PR #576, opened 2019-09-14
autocmd: implement new LoadStarting event

pks-t:
The first autocommand event that is being triggered when loading a
website is the "LoadStarted" event. LoadStarted gets triggered when the
web view has started loading of the page, notably after the initial
request has been sent. Thus, this event comes too late to change
settings that would modify the initial request, like e.g. the user
agent.

Implement a new event LoadStarting that triggers immediately before
performing the initial load and thus before LoadStarted. As WebKit does
not provide any signal for this, we have to manually trigger this event
when executing any load or navigation actions.

This change enables one to modify configuration like the user agent for
certain websites, which was not possible with "LoadStarted".

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -709,6 +709,10 @@ No white space can be used in this list.
 .PD 0
 Events:
 .TP
+.B LoadStarting
+Fired before a new page is going to be opened.
+No data has been sent or received yet, the load may still fail for transport issues.
+.TP
 .B LoadStarted
 Fired if a new page is going to be opened.
 No data has been received yet, the load may still fail for transport issues.
--- a/src/autocmd.c
+++ b/src/autocmd.c
@@ -44,15 +44,16 @@ static struct {
     const char *name;
     guint      bits;
 } events[] = {
-    {"*",                0x00ff},
-    {"LoadStarted",      0x0001},
-    {"LoadCommitted",    0x0002},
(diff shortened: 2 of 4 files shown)


fanglingsu:
@pks-t Thank you for this patch. You are right that the LoadStarted is far to late for request  related changes. I don't like to put the code at that much places to get it working, but if webkit gives us no better hook for it, so we have to work around this by ourselves.
As I can see, the `LoadStarting` event is not triggered when the user click links on pages, or at least not always. Could you please check for a niche place to trigger the event there too?

pks-t:
Yeah, I had been hesitant to add this, too, because of the same reasons. I didn't see your answer at first and don't have any time, but will follow up in the next few days.

pks-t:
Sorry it took me so long, but I've finally fixed the missing
click-event. In fact, I've refactored the patch to make use of
`on_webview_decide_policy`, as it will always get executed
previous to the load. Like this, there's only a single location
instead of multiple ones where we need to trigger the autoload
event.

fanglingsu:
@pks-t Thank you! This works great and looks clean and simple. Great work!

pks-t:
On Mon, Nov 18, 2019 at 12:24:20PM -0800, Daniel Carl wrote:
> @pks-t Thank you! This works great and looks clean and simple. Great work!

Thanks a lot!

PR #580, opened 2019-11-11
Add location setting

fjallarefur:
I added a setting for enabling/disabling location access.
I don't know if there's a better way to do this; feedback is very welcome.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -2008,7 +2008,14 @@ static gboolean on_permission_request(WebKitWebView *webview,
     char *msg = NULL;
 
     if (WEBKIT_IS_GEOLOCATION_PERMISSION_REQUEST(request)) {
-        msg = "request your location";
+        char* geolocation_setting = GET_CHAR(c, "geolocation");
+        if (strcmp(geolocation_setting, "ask") == 0) {
+            msg = "access your location";
+        } else if (strcmp(geolocation_setting, "always") == 0) {
+            return TRUE;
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -2008,7 +2008,12 @@ static gboolean on_permission_request(WebKitWebView *webview,
     char *msg = NULL;
 
     if (WEBKIT_IS_GEOLOCATION_PERMISSION_REQUEST(request)) {
-        msg = "request your location";
+        Setting* location_setting = (Setting*)g_hash_table_lookup(c->config.settings, "location");
+        if (location_setting && location_setting->value.b) {
```
fanglingsu: You could use the macro GET_BOOL(c, "location") to get the configured flag. But I thins it would be better to turn the "location" into "geolocation" and to allow three values like "allow",  "deny" and "ask" or something similar. So the user can allow or deny all geolocation requests or answers each request separate.

fjallarefur: OK, I have updated it to work like that.

fanglingsu:
@fjallarefur Thank you for your work. Your version did not work in case of 'geolocation=always' because the return value of the signal handler is only used to determine if the callback made a decission but not if the request is allowed or not. So I called the `webkit_permission_request_allow` or `webkit_permission_request_deny` in the cases we don't ask the user for the permission decission.

fjallarefur:
I see; I had misunderstood. Thanks for catching that!

PR #583, opened 2019-12-11
Add support for "dark mode"

fjallarefur:
This lets websites use the `prefers-color-scheme` media query to adapt styles according to user preference.
It also affects internal pages like about:blank.
The [Zig docs](https://ziglang.org/documentation/master/) is an example of a website that uses `prefers-color-scheme`, and uses a dark theme if `dark-mode` is enabled.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1143,6 +1143,10 @@ selected.
 .B cursiv-font (string)
 The font family used as the default for content using cursive font.
 .TP
+.B dark-mode (bool)
+Whether to enable dark mode. Websites can use the `prefers-color-scheme' media
+query to adjust styles according to this option.
+.TP
 .B default-charset (string)
 The default text charset used when interpreting content with an unspecified
 charset.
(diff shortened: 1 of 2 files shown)


doc/vimb.1
```diff
@@ -1143,6 +1143,10 @@ selected.
 .B cursiv-font (string)
 The font family used as the default for content using cursive font.
 .TP
+.B dark-mode (bool)
+Whether to enable dark mode. Websites can use the `prefers-color-scheme' media
```
daviwil: Might be worth explicitly saying that this is GTK's dark mode. My initial assumption is that this turns on a custom style sheet in all sites.

daviwil: Though If I had read a little more closely I'd have seen the media query bit :)

fanglingsu:
@fjallarefur Thank you for the patch. I added this media query to my style.css file so I can switch my personal dark-mode. And as a type of test I added a dark version to https://fanglingsu.github.io/vimb/ too.

PR #593, opened 2020-01-09
update arch linux packages

SFort:


fanglingsu:
@SFort Thank you for the patch. I've not seen that the URLs where changed.

PR #603, opened 2020-02-12
fix conditional xembed compilation

raphmaster:
issue #312 is broken. this commit will fix it.


fanglingsu:
@raphmaster Thank you for pointing to this missed feature.

PR #620, opened 2020-06-05
Hinting over an element with no url or source is not an error

mawww:
focusHint returning "OVER:A:" because the hint element did not have
a src or href was treated as an error, and led to the hint keystroke
also being passed down to the page.


fanglingsu:
@mawww Thank you for the patch.

PR #621, opened 2020-06-05
Fix crash in normal_focus_last_active

mawww:
g_variant_get was missing a parameter for the string element of
the variant tuple, pass NULL there as we dont care about that value.


fanglingsu:
@mawww Well spottet. I've never seen this crash on my system. Than you for fixing it.

PR #635, opened 2020-10-17
Create vimb.spec

elagost:
This is a spec file to allow for building of vimb into an rpm. I've tested it and it works on Fedora 32/33/Rawhide.

### DIFF
--- a/vimb.spec
+++ b/vimb.spec
@@ -1,0 +1,36 @@
+Name:           vimb
+Version:        3.6.0
+Release:        1
+Summary:        The vim-like browser
+License:        GPL-3.0-or-later
+Group:          Productivity/Networking/Web/Browsers
+URL:            https://fanglingsu.github.io/vimb/
+Source:         https://github.com/fanglingsu/vimb/archive/%{version}.tar.gz
+BuildRequires:  gtk3-devel, gcc
+BuildRequires:  webkit2gtk3-devel >= 2.20
+
+%global debug_package %{nil}
+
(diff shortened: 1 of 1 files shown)


fanglingsu:
@elagost Thank you for this patch. But I do not want to put build files into the project that relate to a specific packet manager or build system. Could you please write here how a rpm package is build using this spec file. Then can I put this to the web page. But I'll not put it into the code.

elagost:
I appreciate that you do have an opinion on this, and I think it's a good one. Thank you for the response.
The Fedora Project's [documentation on building RPM files](https://docs.fedoraproject.org/en-US/quick-docs/creating-rpm-packages/) does a better job than I could at explaining how to use specfiles.
What I do: download the appropriate version's .tar.gz and the spec file into the same folder, run `fedpkg --release f33 local` (replace 33 with the current version of Fedora) and an .rpm is produced.

AloisMahdal:
@elagost nice job with the specfile.  Have you considered maintaining the package for Fedora?  If that succeeded, all @fanglingsu would need to say on the webpage is that the package is in Fedora.

elagost:
> Have you considered maintaining the package for Fedora?

I've considered it and would like to do it, but it's just not something I've bothered to sit down and do yet. Seems like a large initial time investment to get started.

fanglingsu:
@elagost I close this because packages to not belong into the project. But I there is a package I'd be happy to link it on the web page and the README.

Issue #647, opened 2020-11-12
Unable to view content from the LBRY protocol (lbry.tv / odysee.com)


VictorAlbertsson:
### vimb --bug-info
Version:         3.6.0
WebKit compile:  2.26.2
WebKit run:      2.30.2
GTK compile:     3.24.13
GTK run:         3.24.23
libsoup compile: 2.68.3
libsoup run:     2.72.0
Extension dir:   /usr/lib/vimb

### Steps to reproduce
Go to lbry.tv or odysee.com and click on a video

### Expected behaviour
Video plays

### Actual behaviour
Media player reports: "No compatible source was found for this media"

Any help would be greatly appreciated.

iamleot:
Hello Victor,

Victor Albertsson writes:
> [...]
> ### Steps to reproduce
> Go to lbry.tv or odysee.com and click on a video
>
> ### Expected behaviour
> Video plays
>
> ### Actual behaviour
> Media player reports: "No compatible source was found for this media"
>
> Any help would be greatly appreciated.
> [...]

Probably some gstreamer plugins needed not installed.

I can play videos on lbry.tv here with the following gstreamer plugins
installed (output of `pkg_info' on NetBSD with pkgsrc, but other
package systems probably have similar packages):

 % pkg_info | grep gst
 gstreamer1-1.16.2nb2 Open source multimedia framework
 gst-plugins1-libav-1.16.2nb3 GStreamer libav/ffmpeg plugin
 gst-plugins1-base-1.16.2nb4 Open source multimedia framework - Base plugins
 gst-plugins1-good-1.16.2nb1 Open source multimedia framework - Good plugins
 gst-plugins1-oss-1.16.2nb1 Open source multimedia framework - oss plugin

VictorAlbertsson:
Thank you iamleot, the Arch Linux packages i needed to install where gst-libav and gst-plugins-good.

gst-plugins-good is described by pacman as an optional dependecy for webkit2gtk for media decoding and gst-libav for nonfree media decoding, it would probably be a good idea to add these packages to the README dependencies section since most users will expect media to work out of the box.

Issue #606, opened 2020-03-03
Youtube not loading anymore


AdrienLemaire:
$ vimb --bug-info
Version:         3.6.0
WebKit compile:  2.26.2
WebKit run:      2.26.4
GTK compile:     3.24.13
GTK run:         3.24.14
libsoup compile: 2.68.3
libsoup run:     2.68.4
Extension dir:   /usr/lib/vimb

Youtube worked fine for a while, but suddenly won't load anymore

![youtube](https://user-images.githubusercontent.com/260983/75738522-e2f42880-5d45-11ea-979e-0a1410f43287.png)

I do not have any problem with Youtube on Firefox, and when I look at the network requests there, I couldn't find any call to these sourcemap and map files.
![youtube_firefox](https://user-images.githubusercontent.com/260983/75738854-b2f95500-5d46-11ea-8ab2-141b6cbf1880.png)

AdrienLemaire:
Solved. I broke youtube when I set a user-agent to resolve #601 

Replacing my user-agent with the one you share over there made Youtube work again

```diff
- set user-agent=Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36
+ set user-agent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.1.1 Safari/605.1.15
```

AdrienLemaire:
Or not :(

I can now browse youtube, but no luck playing a video:

![error](https://user-images.githubusercontent.com/260983/75739270-ce189480-5d47-11ea-8d32-6ee33821b415.png)

fanglingsu:
@AdrienLemaire Youtube works fine for me. My User-Agent is actually set to `:set user-agent=Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.0 Safari/605.1.15 vimb/3.6.0-12-ga212b51`. Do you use some adblocking?

AdrienLemaire:
@fanglingsu thanks for the reply.
Using your user-agent didn't help. It's funny that you mention adblocking, because it's when I tried setting up wyebadblock that I realized Youtube was broken.
I did remove the adblock.so and verified that the issue persisted before creating this issue.

Here are the errors thrown by vimb:
```
[...]

PR #648, opened 2020-11-12
Expanded dependencies section

VictorAlbertsson:
These additions to the README clarify which packages need to be installed to get media to work in vimb and thus solves #647 and also #606

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,9 @@ the project page of [Vimb][].
 
 - gtk+-3.0
 - webkit2gtk-4.0 >= 2.20.x
+- gstreamer
+- gst-libav
+- gst-plugins-good
 
 ## Install
 
--- a/README.md
+++ b/README.md
@@ -39,9 +39,8 @@ the project page of [Vimb][].
 
 - gtk+-3.0
 - webkit2gtk-4.0 >= 2.20.x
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,8 @@ the project page of [Vimb][].
 
 - gtk+-3.0
 - webkit2gtk-4.0 >= 2.20.x
+- gst-libav (for media decoding)
+- gst-plugins-good (for nonfree media decoding)
```
iamleot, line 43: Strictly speaking both are optional and it would be nice to says that in some ways, maybe adjusting the wording to maybe just one item, e.g.:

```
 - gst-libav, gst-plugins-good (optional, needed for media playback)
```

(I also do not think that gst-plugins-good are useful for "nonfree" media only.)

iamleot (changes requested):

README.md
VictorAlbertsson: Added 'optional' to description also mentioned that 'gst-plugins-good' might be needed for other things as well.

VictorAlbertsson:
If no further changes are desired please approve this pull request.

fanglingsu:
@VictorAlbertsson Thank you for the patch.

PR #662, opened 2021-02-27
add kill to hint mode

howeyc:
very useful to kill ads or anything you don't want to see


fanglingsu:
@howeyc Thank you this is a nice litte patch - even if it is hard to see which div is the right one to remove.

PR #669, opened 2021-04-15
Add setting for toggling javascript markup option

Sadoon-AlBader:
Older systems like ppc32 struggle with WebKit's JavaScript engine, having this setting turned off enables those systems to access most websites with limited JavaScript functionality instead of simply crashing WebKit on any JS website.

Was tested on Gentoo and ArchPower, works great.

PS: Thank you for creating this browser, it's refreshing to use such a fast browser with a small and simple codebase that people can look into and contribute to!

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1538,6 +1538,12 @@ This fills the inputbox with the prefilled download command and replaces
 Whether to enable the XSS auditor.
 This feature filters some kinds of reflective XSS attacks on vulnerable web
 sites.
+.PD
+.RE
+.TP
+.B javascript-enable-markup (bool)
+Whether JavaScript markup is enabled.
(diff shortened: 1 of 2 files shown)


fanglingsu:
@Sadoon-AlBader Thank you for your investigation. Could you please document this new setting in the man page `doc/vimb.1` too. And we currently support webkit versions down to 2.20.x and this setting is only available from 2.24 on, so you should check the webkit version during compilation like this https://github.com/fanglingsu/vimb/blob/0b0060d95885f280bc635ea24d73078880951b29/src/setting.c#L109-L112

Sadoon-AlBader:
Just fixed both issues, sorry for taking so long :)
I'm just now starting to use vimb as my main browser.
Hope these commits make everything complete.

fanglingsu:
@Sadoon-AlBader Thank you!

PR #680, opened 2021-09-16
Fix scroll percentage on ppc64le, ppc64, ppc & armv7l (possibly others)

Sadoon-AlBader:
Version:         3.6.0-39-ged7d365
WebKit compile:  2.32.3
WebKit run:      2.32.3
GTK compile:     3.24.24
GTK run:         3.24.24
libsoup compile: 2.72.0
libsoup run:     2.72.0
Extension dir:   /home/$USER/.local/lib/vimb

(x86, x86_64, ppc64le)

Version:         3.6.0-34-g652f411
WebKit compile:  2.32.3
WebKit run:      2.32.3
GTK compile:     3.24.30
GTK run:         3.24.30
libsoup compile: 2.74.0
libsoup run:     2.74.0
Extension dir:   /home/$USER/.local/lib/vimb

(ppc64, ppc)

Version:         3.6.0-37-g837ff36
WebKit compile:  2.32.3
WebKit run:      2.32.3
GTK compile:     3.24.5
GTK run:         3.24.5
libsoup compile: 2.64.2
libsoup run:     2.64.2
Extension dir:   /home/pi/.local//lib/vimb

(rpi4)

### Steps to reproduce
Use the browser

### Expected behaviour
Correct scroll percentage shown

### Actual behaviour
Different behavior accross architectures:
(x86, x86_64, and ppc64le using Devuan Linux Chimaera; ppc64 and ppc using Debian Sid; Rpi4 using Raspbian (Debian 10) )

On x86 and x86_64, the values show as expected.

on ppc64le, ppc64, ppc and armv7l, different outcomes are observed, I chose to display the value in hexadecimal, making it easier to find the specific bug.

ppc, ppc64: Lower 2 bytes are always set to 0x0000, upper 2 bytes contain the percentage of scrolling.

ppc64le: Upper two bytes are set to random numbers on each page, lower two bytes contain the percentage of scrolling.

armv7l: Upper two bytes are set to 0xABC0 always, lower bytes are the percentage.

The fix: Using a guint16 instead of guint seems to completely fix the issue without introducing bugs on any architecture.

I can't figure out for the life of me why this is happening, but I'm going to investigate libwegkit2gtk later because I am very sure it's an issue with the library itself.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -632,6 +632,15 @@ void vb_statusbar_update(Client *c)
 
     statusbar_update_downloads(c, status);
 
+		/**
+		* These architectures have different kinds of issues with scroll 
+		* percentage, this is a somewhat clean fix that doesn't affect	others.
+		*/
+#if defined(_ARCH_PPC64) || defined(_ARCH_PPC) | defined(_ARCH_ARM)
+		/* force the scroll percent to be 16-bit */
+		c->state.scroll_percent = * (guint16*) ( &c->state.scroll_percent );
+#endif
+
     /* show the scroll status */
     if (c->state.scroll_max == 0) {
         g_string_append(status, " All");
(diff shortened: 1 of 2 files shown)


Sadoon-AlBader:
It seemed that the fix was only working on the rpi4, ppc, and ppc64le, and ppc64 was still having an issue.
I decided to clean up the patch so that it treats the scroll percentage as a 16-bit uint only on affected architectures. Overall I think it's much cleaner.
I know "pointer manipulation" and whatnot but it makes it work both on big endian and little endian systems this way.

fanglingsu:
@Sadoon-AlBader Thank you for you investigation an for the patch!

PR #681, opened 2021-09-18
make invalid command warnings more useful

Sadoon-AlBader:
Added a simple print of line numbers when printing a warning of invalid commands in the config file. Saved me some headache finding that line with only spaces in it.
I considered fixing it so that it doesn't warn on lines with spaces and tabs and considers them empty instead, but it might add more complexity than necessary.


fanglingsu:
@Sadoon-AlBader Pretty nice idea to make dearching for wrong syntax in config easier.

PR #683, opened 2021-09-21
New feature: show some settings on statusbar (js, cookies, html5-localstorage)

Sadoon-AlBader:
I added a new feature that shows the status of JavaScript, cookies, and HTML5 LocalStorage on the right hand side of the statusbar, next to the scroll percentage.
This can be toggled by the setting `status-bar-show-settings`

Before:

statusbar_before

After:

statusbar_after

I also fixed a small mistake I did before with the man file, "javascript-enable-markup" was not in alphabetical order so I fixed that.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -632,6 +632,18 @@ void vb_statusbar_update(Client *c)
 
     statusbar_update_downloads(c, status);
 
+    if ( c->config.statusbar_show_settings ) {
+        /* show js, cookies, and local storage status */
+
+        Setting *js = g_hash_table_lookup(c->config.settings, "scripts");
+        Setting *cookies = g_hash_table_lookup(c->config.settings, "cookie-accept");
+        Setting *local_storage = g_hash_table_lookup(c->config.settings, "html5-local-storage");
+
+        g_string_append_printf(status, " js: %s |", js->value.b ? "on" : "off");
+        g_string_append_printf(status, " cookies: %s |", cookies->value.s);
+        g_string_append_printf(status, " storage: %s", local_storage->value.b ? "on" : "off");
+    }
+
     /* show the scroll status */
     if (c->state.scroll_max == 0) {
         g_string_append(status, " All");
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -632,6 +632,18 @@ void vb_statusbar_update(Client *c)
 
     statusbar_update_downloads(c, status);
 
+    if ( c->config.statusbar_show_settings ) {
+        /* show js, cookies, and local storage status */
+
+        Setting *js = g_hash_table_lookup(c->config.settings, "scripts");
+        Setting *cookies = g_hash_table_lookup(c->config.settings, "cookie-accept");
+        Setting *local_storage = g_hash_table_lookup(c->config.settings, "html5-local-storage");
```
fanglingsu, line 640: @Sadoon-AlBader Could you please use the macros out of src/main.h to read the config settings. `GET_BOOL(c, "scripts")`, `GET_CHAR(c, "cookie-accept")` and `GET_BOOL(c, "html5-local-storage")`

src/main.c
```diff
@@ -632,6 +632,18 @@ void vb_statusbar_update(Client *c)
...
+
+        g_string_append_printf(status, " js: %s |", js->value.b ? "on" : "off");
+        g_string_append_printf(status, " cookies: %s |", cookies->value.s);
+        g_string_append_printf(status, " storage: %s", local_storage->value.b ? "on" : "off");
```
fanglingsu, line 644: Might it possible to allow the user to define these values by config.h file. I Think this is a nice feature to see which settings are set. But users might want so se other values here.
I think it would be nice to define the settings to be read and the visualization to compile time. But I'm not sure if this would be easy possible with some kind of array of structs ore something similar.

Sadoon-AlBader: Hmm it is possbile only if we limit ourselves to one kind of value: ints, strings, or booleans.
Because as you can see, cookies have three possible options, while the others are just booleans.
I'll think of a way to handle that and see how it goes.

src/main.c
Sadoon-AlBader: I wish I had seen those before, wasted a lot of time trying to come up with my own way xD

Sadoon-AlBader: Alright, done. Seems to work well.

src/main.c
Sadoon-AlBader: Alright I got it done, let me know if you like the change so I can update the man doc :)

PR #685, opened 2021-09-26
Add ability to map PageUp and PageDown

Sadoon-AlBader:
Webkit2Gtk 2.34.0 is out and added a (to me) annoying feature that "smooth scrolls" on PageUp and PageDown even when smooth scrolling is disabled. It's very buggy (at least for me on ppc64le and ppc64, haven't tested others)

This allows us to map PageUp and PageDown to other keys (like <C-b> and <C-f> in this case) to scroll full pages without animations, works much better, and opens up some possibilities for other users.


fanglingsu:
@Sadoon-AlBader This looks good on a first view over it. I'l test and merge it soon. Thank you!

PR #693, opened 2021-12-14
Add option for WebKit Intelligent Tracking Prevention

illfygli:
Please let me know if I got the version checks right.
I could only find examples of checking the runtime version in vimb, but the ITP functions were [introduced in 2.30](https://webkitgtk.org/reference/webkit2gtk/unstable/WebKitWebsiteDataManager.html#webkit-website-data-manager-set-itp-enabled) and I guess compilation would fail if I don't check the compile-time version.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1364,6 +1364,9 @@ CSS style applied to the inputbox in normal state.
 .B input-error-css (string)
 CSS style applied to the inputbox in case of displayed error.
 .TP
+.B intelligent-tracking-prevention (bool)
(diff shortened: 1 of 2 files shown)


fanglingsu:
@illfygli You should use `WEBKIT_CHECK_VERSION(2, 30, 0)` which evaluates to true is the version is at least 2.30.0 https://webkitgtk.org/reference/webkit2gtk/stable/webkit2gtk-4.0-WebKitVersion.html#WEBKIT-CHECK-VERSION:CAPS

illfygli:
Thanks, updated.

fanglingsu:
@illfygli Thank you!

PR #694, opened 2021-12-23
Add hints for <summary> elements

illfygli:
The [`<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary) element has an implicit ARIA role of `button`, and is typically used to toggle the visibility of content within a `<details>` element.
They are used on Sourcehut among other websites and it's nice to be able to toggle them easily.
There's a demo on the [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary).


fanglingsu:
@illfygli Thank you! I've never seen this before.

PR #697, opened 2021-12-29
Set z-index on hint container to max value

illfygli:
There was a GDPR consent dialog that made itself inaccessible by setting `z-index` to the maximum `i32` value `2147483647`, thereby covering the hint container and hints.
This seems to be the largest value supported in practice.
Setting the hint container's `z-index` to the same value puts in on top, since it is also last in the DOM.


fanglingsu:
@illfygli It sounds strange to mee to have such an high z-index - but as this seems to be required I merge this.

PR #706, opened 2022-01-26
Add the shellopen command

doxnit:
Adds a new feature - the **shellopen** command. **Shellopen** takes a shell command as an argument and opens an URI returned by it

### DIFF
--- a/src/ex.c
+++ b/src/ex.c
@@ -81,6 +81,7 @@ typedef enum {
     EX_SCR,
     EX_SET,
     EX_SHELLCMD,
+    EX_SHELLOPEN,
     EX_SOURCE,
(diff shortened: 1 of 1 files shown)


cdlscpmv:
Some time ago I was thinking about a similar feature. In my opinion, it would be more useful to be able to return an Ex command from a user script instead of just an URI and have vimb execute it (similar to what Vim's `:exe` does). This would provide a quite useful IPC mechanism without any unix domain sockets or named pipes. 

With this feature, opening a new URI from a user script is just a matter of prefixing the returned URI with `open` or `tabopen`.

doxnit:
@cdlscpmv, this is very good idea, I'll try to implement it

PR #707, opened 2022-01-28
Add shellex command

doxnit:
Adds a new feature - the **shellex** command. **Shellex** takes a shell command as an argument, reads its output line by line, and executes those lines as vimb commands.

Thanks @cdlscpmv for idea.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -477,7 +477,7 @@ normal
 .IP -
 open, tabopen
 .IP -
-shellcmd
+shellcmd, shellex
 .PD
 .
 .SS Command Line Editing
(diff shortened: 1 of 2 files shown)


fanglingsu:
@doxnit Thank you for the patch and sorry for the late response. For which tasks do you use the :shellex? Yould you please update the manpage under doc/vimb.1.
Could you please reuse existing code of `ex_shellcmd`. Your `ex_shellex` looks nearly the same so I assume you could restructure the code so that it can handle :shell and :shellex withou replicating the code.

doxnit:
@fanglingsu Sorry for the very late response. I added the shellex description with an example to the manpage file. I also extracted VIMB_SELECTION related code to a function to reduce code duplication.
You can use the :shellex for many things. The idea behind :shellex is to have ability to run ex commands from other programs (kind of opposite of the :shellcmd). For example, I have keybind to pipe history/bookmarks into fzf and then open choosen url in the current window. I also use it to inject env variables into other ex commands. For example, I have 'shellex echo ":set home-page=file://$XDG_CONFIG_HOME/vimb/home.html"' in my config (should we set EX_FLAG_EXP for the ex_set?). You can also use it to make a menu in which you can select some javascript file and eval it.
I also fixed memory leakage in the shellcmd (run :shellcmd du / and check memory usage before and after and you will notice difference), removed duplicate of :shellcmd! in the manpage and fixed manpage markup a little.
English is not my native language so it would be nice if someone would check and correct manpage changes.
Thank you!

PR #710, opened 2022-05-08
README.md: update Gentoo overlays: remove hsoft, add jjakob

jjakob:
https://github.com/hsoft is 404 so remove it.
There doesn't seem to be an overlay named "hsoft" in the gentoo overlay
list either.
Add https://github.com/jjakob which has an updated ebuild from the
latest git commit, that is not a live ebuild.

Live ebuilds (version 9999, using git-r3) are considered insecure since there's no verification of the downloaded source. Using a versioned ebuild with SRC_URI is secure as the checksum of the source archive is saved in the Manifest.
Since the tharvik overlay only has 3.6.0 and 9999 ebuilds, I was stuck with either an old version, or an insecure ebuild, so I created an ebuild with the latest git commit as the source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ the project page of [Vimb][].
 ## Packages
 
 - Arch Linux: [community/vimb][], [aur/vimb-git][], [aur/vimb-gtk2][]
-- Gentoo: [gentoo-git][], [gentoo][]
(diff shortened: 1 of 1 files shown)


fanglingsu:
@jjakob Thank you for the patch and for providing an ebuild.

PR #711, opened 2022-07-09
Read-it-Later Queue Improvements

VictoriaLacroix:
I've been finding the Read-it-Later Queue to be very useful and at this point it has almost completely replaced the need to have dozens of tabs. I've made a few changes to help make it easier to use for certain use cases, offering here in the event that the changes are desirable.

First is to allow bookmark completion with the `:qpush` and `:qunshift` commands. I use this feature a lot for reading sites that update frequently (mainly Twitter feeds as I don't have an account).

Second is new key combinations to queuing items. 'q' and 'w' which prompt the user to `:qpush` and `:qunshift` respectively. 'Q' and 'W' do the same but also include the current URI. I find the latter two keys helpful for when I'm consuming my queue. 'W' can un-pop the last item in case I want to keep it. 'Q' is very helpful in case I ever reach an item I would like to keep queued but don't want to consume just yet.

### DIFF
--- a/src/ex.c
+++ b/src/ex.c
@@ -1301,6 +1301,8 @@ static gboolean complete(Client *c, short direction)
             switch (arg->code) {
                 case EX_OPEN:
                 case EX_TABOPEN:
+                case EX_QPUSH:
+                case EX_QUNSHIFT:
                     sort = FALSE;
                     if (*token == '!') {
                         found = bookmark_fill_completion(store, token + 1);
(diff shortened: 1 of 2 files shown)


fanglingsu:
@VictoriaLacroix Thank you for the patch. The completion for the queue push and unshift ist a nice feature. But I can't see the need for the q,w and Q and W bindings. If I read it right - that are alls things you could also do by remapping keys.
```
:nnoremap Q :qpush <C-R>%
:nnoremap q :qpush<Space>
```
The `<C-R>%` is filled by the register "%" which holds the current url.

fanglingsu:
@VictoriaLacroix I've merged the first commit into master. But the additional key mappings are not required for me and other users. And could be easily be applied with current code.

PR #722, opened 2022-09-24
Improve certain default settings

amarakon:
I made three changes:

1. Set the default download path to `~/Downloads`, this is more common and no one likes having their home directory cluttered
2. Enable `input-autohide` by default because most of the time it will just be empty space. Zathura also does this by default.
3. Enable `incsearch` by default to preview while searching. This is enabled in Neovim by default and it makes searching much easier.


fanglingsu:
Thank you for your improvements.

PR #725, opened 2022-12-21
main: fix compilation with FEATURE_NO_XEMBED

pks-t:
When FEATURE_NO_XEMBED is defined then we don't support using the Xembed interface, which will cause us to set the `VIMB_XID` envvar. With 55fb9cc (Add new env VIMB_WIN_ID to hold the gtk window id., 2021-01-25) a related feature was added that causes us to expose the Gtk window ID via the `VIMB_WIN_ID` environment variable. This feature also depends on X-specific information because it uses `GDK_WINDOW_XID()` to derive the window ID. This change thus broke Wayland-only environments.

Fix compilation by extending the `ifdef` to guard the logic that both sets the `VIMB_XID` and the `VIMB_WIN_ID`.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -833,10 +833,10 @@ static void client_show(WebKitWebView *webview, Client *c)
 
     gtk_widget_show_all(c->window);
 
+#ifndef FEATURE_NO_XEMBED
     char *wid;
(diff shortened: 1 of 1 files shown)


fanglingsu:
@pks-t Thank you for this contribution.

PR #728, opened 2023-02-05
Add global marks

### LINKED ISSUE #598: Permanent marks and global marks
Related: #355 

Hi and thanks for making this cute browser. Just found it yesterday, and looking forward to becoming proficient with it.

One thing that I'm missing from [Vimium](https://github.com/philc/vimium) are better marks:

* using backtick ` as well as single quote ' to jump to a mark
* lowercase letters register local marks; uppercase letters register global marks
* marks are remembered permanently (either repeat mX to toggle mark or :delm ?)

Is that not implemented in vimb because you consider bookmarks to be enough?

Example: Access vimb with mark/tag V:
* `mV` then `'V`
* `:bma V` then `:open !V`

Note that marks not only save the url but also the location in the page, which can be very handy when you want to jump to a specific paragraph in a very long webpage. Since vimb clears local tags after leaving the page, it can be tedious to find back the right location in a page if some search pattern to get there is repeated multiple times in that page.

BioBox:
Fixes #598. I've done some testing on it and did my best to conform to the contributing guidelines, but I felt that using a global variable declared immediately before the callback definition was the right call here.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -225,8 +225,15 @@ Set a page mark {\fIa-z\fP} at the current position on the page.
 Such set marks are only available on the current page;
 if the page is left, all marks will be removed.
 .TP
+.BI m{ A-Z }
+Set a page mark {\fIA-Z\fP} at the current position on the page.
+These marks are permanent across any page.
+.TP
 .BI '{ a-z }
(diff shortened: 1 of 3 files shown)


fanglingsu:
@BioBox Thank you for the patch. That's a really cool feature.

PR #744, opened 2023-06-29
Document how the config file works.

falsifian:
Reading the man page, I didn't realize I could put ex commands in the `config` file. The description made me think it was just for the specific things listed (webkit settings, gui styles, keybindings). (And it also left me wondering exactly what the format was.)

Thanks to Konst on the mailing list for telling me I can put any commands in it.

If you'd prefer, I can try to edit the description so it also talks about styles and keybindings, but to me this version seems simple and straightforward.

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1573,7 +1573,7 @@ this subdirectory.
 .PD 0
 .TP
 .I config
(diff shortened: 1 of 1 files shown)


fanglingsu:
@falsifian Thank you for your contribution.

PR #752, opened 2023-08-30
Escape special keys in mappings using backslash

### LINKED ISSUE #753: Add a way to escape special keys in mappings
I would like a way of escaping special keys like `<C-R>` in mapping. My particular use case is makeing a mapping that opens a hinted URI in another browser like Firefox. Consider the mapping
```
:nnoremap ;f :set x-hint-command=:sh! firefox '<C-R>;'<CR>;x
```
There are two issues with this

 1. The key sequence `<C-R>;` is executed as soon as `;f` is pressed.
 2. This means that `x-hint-command=:sh! firefox ''`. Notice that there's no way to get `<CR>` at the end of this, so after a URI is hinted we're left with `:sh! firefox ''` in the command line and have to press enter manually to execute the command.

In PR #752 I add the ability to escape with backslash, the result being that the mapping
```
:nnoremap ;f :set x-hint-command=:sh! firefox '\<C-R>;'\<CR><CR>;x
```
works as intended. The syntax `\<C-R>` causes this string to be interpreted as the key sequence `<`, `C`, `-`, `R`, `>`. The PR also adds a `<Bslash>` special key so that `<Bslash><C-R>` is backslash followed by CTRL-R.

Two questions that need to be answered:

 1. Syntax. Is this syntax ok? Would another syntax be more desirable? Another idea I had was to make `<<>` into a literal `<` character, so that instead of `\<C-R>` you would type `<<>C-R>`. This would make it so that we don't need to add a `<Bslash>` special key.
 2. Is there already another way to accomplish the same things a new escape syntax would accomplish?

ndtodoroff:
* Interpret the syntax `\<C-R>` as representing the literal key sequence `<`, `C`, `-`, `R`, `>`.
* Add a `<Bslash>` special key so that e.g. the syntax `<Bslash><C-R>` is interpreted as the key sequence `\` followed by CTRL-R.

My personal use case is the following mapping:
```
:nnoremap ;f :set x-hint-command=:sh! firefox '\<C-R>;'\<CR><CR>;x
```
this mapping starts hint mode and then opens the hinted URI in Firefox.

Closes #753

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+* Allow special keys to be escaped in mappings using `\`. For example, `\<C-R>`
+  represents the key sequence `<`, `C`, `-`, `R`, `>`.
+* Add `<Bslash>` as an alias for `\` in mappings. For example, `<Bslash><C-R>`
+  represents the key sequence `\` followed by CTRL-R.
 
 ## [3.7.0] - 2023-06-19
 ### Added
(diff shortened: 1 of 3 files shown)


fanglingsu:
@ndtodoroff Thank you for this pretty useful improvement.

PR #772, opened 2024-08-29
Appends to registers if capitals are used

### LINKED ISSUE #771: Use capital register names to append
### Steps to reproduce
1. `"A;y`, then select a hint
1. `:reg`, then note that register `a` is empty

### Expected behaviour
Using capital letter of a register should append yanked content to the register (as per vim's convention).

### Actual behaviour
Capital letters for registers do not constitute a regular command.

machinedgod:
By default -  it separates entries by newline.

Closes #771.

If you'd like to merge this, let me know if anything about the code isn't to your liking and I'll fix it. If not, please just let me know and reject the request :)


machinedgod:
Changing PR to go from separate branch (I've added `:register` alignment for printing)

PR #781, opened 2025-04-10
Allow overriding of GDK window color

### LINKED ISSUE #594: feature request: avoid white views when dark-mode is on.
Vimb 3.6.0 introduces the dark-mode, which is great! However, when starting vimb with a page, it first shows a grey background, then white, then dark.

Reproduce: 
Execute "vimb https://codevoid.de" (or any other page with a dark background).
It also happens on `:tabopen`.

I'm not sure if this can be influenced (because webkit...). But it would be great it these initial views could be a dark grey when dark-mode is on.

Best Regards,
Sdk

MichaelMackus:
This prevents a white flash for example when using dark mode.

Fixes #594

### DIFF
--- a/src/config.def.h
+++ b/src/config.def.h
@@ -45,6 +45,8 @@
 #define SHOWCMD_LEN                 10
 /* css applied to the gui elements regardless of user's settings */
 #define GUI_STYLE_CSS_BASE          "#input text{background-color:inherit;color:inherit;caret-color:@color;font:inherit;}"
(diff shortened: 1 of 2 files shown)


MichaelMackus:
I'm sure there's a way to parse the setting for dark mode, but this allows flexibility in choosing the specific window color the user wants (without parsing the user's CSS file)

fanglingsu:
@MichaelMackus Than you for this patch.

Issue #762, opened 2024-03-01
Slow text search


BebeSparkelSparkel:
### Steps to reproduce

Go to  https://hackage.haskell.org/packages/ and type in `/data` to search for text containing `data`.

### Expected behaviour

I expect it to find the text faster than I can find it with my eyes.

### Actual behaviour

Takes multiple seconds to return any results.

### Posible solution

Perhaps incrementally return results as found instead of finding all results and then returning them.

doxnit:
Resolved by #784. I think we can close this issue

PR #784, opened 2025-08-28
Improve search performance

doxnit:
If incsearch is enabled Vimb does searching on each keystroke. It looks like WebKit does not abort previous searches. It leads to very poor search performance on huge documents like in the #762. My solution to this problem is to limit each incsearch up to INCSEARCH_MATCHES_LIMIT matches. Also the "webkit_find_controller_count_matches" call and the "counted-matches" callback were replaced by 2 new callbacks (~1.5x faster on my machine).


fanglingsu:
@doxnit Thank you!

PR #787, opened 2025-09-08
Add FEATURE_HISTORY_WITHOUT_HOME_PAGE

doxnit:
Some users may prefer their history file isn't flooded by home-page uri records

### DIFF
--- a/src/config.def.h
+++ b/src/config.def.h
@@ -28,6 +28,8 @@
 #define FEATURE_QUEUE
 /* disable X window embedding */
 /* #define FEATURE_NO_XEMBED */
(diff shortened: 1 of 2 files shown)


doxnit:
@cdlscpmv Yes I do

cdlscpmv:
I think other people using a local home page experience a similar problem. Perhaps, it would be better to add an additional match for the `file:///` URLs and remove the preprocessor condition. @fanglingsu, what do you think?

doxnit:
I created new PR (#788) with histignore setting. If new PR will be merged then we can close this one.

PR #795, opened 2026-02-19
Fix hard-codes pkg-config

mati75:
vimb fails to cross-build due to hard-coding pkg-config in config.mk. The right arch-triplet prefixed pkg-config is passed by default in cross builds which this should use.

See debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110406


fanglingsu:
@mati75 Thank you for the patch. Could you please check how we can avaoid the clang error about recursive variable reference?
```
Run make options && make -j test
  make options && make -j test
  shell: /usr/bin/bash -e {0}
  env:
    CC: clang
config.mk:15: *** Recursive variable 'PKG_CONFIG' references itself (eventually).  Stop
```

fanglingsu:
@mati75 Thank you for your work!

Issue #809, opened 2026-05-21
strict-ssl=off doesn't work


jjakob:
When trying to access a server with a self-signed certificate, strict-ssl=off used to work, now it doesn't, it always shows "Unacceptable TLS certificate".

```
Version:         4.0.0_alpha_p20260328+git6e826758
WebKit compile:  2.50.5
WebKit run:      2.50.5
GTK compile:     4.18.6
GTK run:         4.18.6
libsoup compile: 3.6.5
libsoup run:     3.6.5
Extension dir:   /usr/lib/vimb
```

PR #810, opened 2026-06-07
Fix webviews don't use session

### LINKED ISSUE #807: Cookies lost after closing
When I close and reopen vimb, all cookies (and consequently all logins) saved in the previous session are lost.

doxnit:
Webviews should be created with the network-session parameter. Otherwise, Vimb doesn't use a cookies file. Resolves #807


doxnit:
Probably it also resolves #809

cdlscpmv:
Can confirm that this also fixes #809.

PR #812, opened 2026-06-10
Fix downloads

### LINKED ISSUE #804: Downloads don't work in master branch
Downloads don't work. The status bar changes as if the file is starting to be downloaded but the progress bar never moves. Nothing is logged in the console or anywhere.

```
Version:         4.0.0_alpha_p20260328+git6e826758
WebKit compile:  2.50.5
WebKit run:      2.50.5
GTK compile:     4.18.6
GTK run:         4.18.6
libsoup compile: 3.6.5
libsoup run:     3.6.5
Extension dir:   /usr/lib/vimb
```

doxnit:
Fix downloads. Resolves #804

### DIFF
--- a/doc/vimb.1
+++ b/doc/vimb.1
@@ -1264,10 +1264,11 @@ Setting value of 'download-path' which would be used normally for downloads.
 .TP
 .B download-path (string)
 Path to the default download directory.
-If no download directory is set, download will be written into current
-directory.
-The following pattern will be expanded if the download is
-started '~/', '~user', '$VAR' and '${VAR}'.
+If no download directory is set, download will be written into
+$HOME/Downloads. It must be absolute path. You can take advantage of the
+:shellex command to expand relative path or environment variables:
+.IP
+:shellex sh -c "echo -n :set download-path=; realpath ~/downloads/"
 .TP
 .B download-use-external (bool)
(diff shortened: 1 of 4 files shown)


cdlscpmv:
Thanks for you work, @doxnit! The fix does work.

However, `XDG_DOWNLOAD_DIR` is not part of [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/), so it is not a good idea to put it in code. The user can assign `download-path` to it using the `:shellex` command if they want to.

doxnit:
Thank you, @cdlscpmv! I mistakenly thought that XDG_DOWNLOAD_DIR is part of the specification. I should have double checked this. I have pushed changes so it should be fine now

cdlscpmv:
@doxnit, thanks for the quick update.

Overall, I think we shouldn't touch the logic of how `download-path` behaves, because that would break backward compatibilty (and it is good as it is now, in my opinion). So I would suggest to keep only the changes that fix the actual problem.

fanglingsu (commented):
@doxnit could you please use glib functions in get_default_download_path istead of memcpy and malloc calls.

doxnit:
@fanglingsu Done

fanglingsu:
@doxnit Thank you

PR #815, opened 2026-06-20
Fix input mode bugs

doxnit:
Fixes several bugs:
- hitting 'i' do not focus first focusable element (it should do this if there is no last focused element acording to the manpage);
- Vimb doesn't leave input mode on new page. Open google.com, type something in the search box and hit Enter. Vimb will be in the input mode on a results page and will behave like it's in the pass-through mode.
- Vimb doesn't handle focus changes before document has been loaded. Open google.com and you will see that webkit set focus to the search box (because the input has autofocus attribute) but Vimb is still in the normal mode.


fanglingsu:
@doxnit Thank you for your investigations and for providing these fixes. Great work!

